PR #23, opened 2014-10-13
Python example using pyopengl

marblestation:
I have implemented a python example that shows how pyopengl can be used to visualize the simulation, similar to what can be seen for the C examples. It requires pyopengl (pip install pyopengl).


hannorein:
It has been a few years, but I'll close this now. I think going forward the only reasonable real-time visualization options are OpenGL for the c-based version and whatever works in jupyter notebooks from the python side.

PR #29, opened 2015-01-07
Megno

dtamayo:
I made the change we talked about to set delta for the central body such that the center of mass's position and velocity are unchanged in the shadow system.


hannorein:
I think you want to send the pull request to the MEGNO branch.

dtamayo:
So close!  The good news is I think I actually understand how pull requests work now :)  I made a new pull request.

Thanks,

Dan

PR #30, opened 2015-01-07
edited integrator_megno_init so that the center of mass pos. and vel. ar...

dtamayo:
...e unchanged in the shadow system


dtamayo:
Hi Hanno,

Great, thanks for the feedback.  Sure, let's talk about it in person,

Dan

PR #31, opened 2015-01-16
init_planet and p2orbit functionality, plus simple python 2/3 compatibility changes to rebound.py

dtamayo:


python_examples/rebound.py
```diff
@@ -1,4 +1,9 @@
 from ctypes import *
+import numpy
```
hannorein: Hm. I don't like that it uses numpy. I think `math` should be able to do the same.

python_examples/rebound.py
```diff
@@ -1,4 +1,9 @@
...
+import sys
```
hannorein: Similarly, I think we can do this without the package. How about

    try:
        xrange:
    except NameError:
        xrange = range

python_examples/rebound.py
```diff
@@ -127,4 +149,158 @@ def step():
...
+                Omega=None, # longitude of ascending node
+                mass=None,  # planet mass
+                M=None):    # mean anomaly
+    '''Returns a particle structure to add into rebound for a given set of
+        orbital elements. 'star' and 'a' are required, others will default to zero
+        (with a warning). Only f OR M should be passed to specify the location of
+        the particle on its orbit.'''
+    
+    # first check for valid inputs and give warnings on defaults
+    if e == None: e = 0.; print('e not passed, setting to 0... (circular)')
```
hannorein: I don't think I would print out these warning messages. It seems pretty obvious that if you don't pass `e`, it will be 0 (rather than 0.2348764 or 1). Anyone still not sure what it does can look up your documentation of the function.

python_examples/rebound.py
```diff
@@ -127,4 +149,158 @@ def step():
...
+        E = E - F/(1.0-e*numpy.cos(E))
+        F = E - e*numpy.sin(E) - M
+        if numpy.fabs(F)<1e-16:
+            break
+    E = mod2pi(E)
+    return E
+
+def init_planet(star,       # central object (rebound.Particle object)
+                a,          # semimajor axis
+                e=None,     # eccentricity
```
hannorein: If there are no warning messages, one could change this line to 

    e=0.

And get rid of the extra line below.

python_examples/rebound.py
```diff
@@ -127,4 +149,158 @@ def step():
...
+    if not(0. <= inc <= numpy.pi): raise ValueError('inc must be in range [0,pi]')
+    
+    if f == None: # need to calculate f
+        E = get_E(e,M)
+        f = mod2pi(2.*numpy.arctan(numpy.sqrt((1. + e)/(1. - e))*numpy.tan(0.5*E)))
+    
+    r = a*(1-e**2)/(1 + e*numpy.cos(f))
+    
+    # Murray & Dermott Eq. 2.122
+    _x  = star.x + r*(numpy.cos(Omega)*numpy.cos(omega+f) - numpy.sin(Omega)*numpy.sin(omega+f)*numpy.cos(inc))
```
hannorein: This ok, but inefficient. It might be worth caching all the `cos` and `sin` functions rather than calculating them multiple times. They can take a long time to compute.

hannorein: Purely ecstatic, but I don't think we need the underscores here `_x`.

python_examples/rebound.py
```diff
@@ -127,4 +149,158 @@ def step():
...
+    _vz = star.vz + (n*a/numpy.sqrt(1-e*e))*((e+numpy.cos(f))*numpy.cos(omega)*numpy.sin(inc) - numpy.sin(f)*numpy.sin(inc)*numpy.sin(omega))
+    
+    return Particle(m  = mass, x = _x, y = _y, z = _z, vx = _vx, vy = _vy, vz = _vz)
+
+def p2orbit(p, star):
+    '''Takes 2 Particle objects (one for object and one for central body,
+        which hold Cartesian elements) and returns a Orbit object
+        with all the orbital elements'''
+    
+    if star.m <= 1.e-30: raise ValueError('Star has no mass.')
```
hannorein: Why the magic number `1e-30`? The smallest floating point number in double precision is ~`1e-308`. I'd suggest using something closer to that value.

python_examples/rebound.py
```diff
@@ -127,4 +149,158 @@ def step():
...
+    p.y -= star.y
+    p.z -= star.z
+    p.vx -= star.vx
+    p.vy -= star.vy
+    p.vz -= star.vz
+    h0 = (p.y*p.vz - p.z*p.vy)                  # angular momentum vector
+    h1 = (p.z*p.vx - p.x*p.vz)
+    h2 = (p.x*p.vy - p.y*p.vx)
+    o.h = numpy.sqrt ( h0*h0 + h1*h1 + h2*h2 )  # abs value of angular moment
+    if o.h <= 1.e-30: raise ValueError('Particle orbit is radial.')
```
hannorein: Similar issue here, it's very good to have a failsafe in here, but it should probably be significantly smaller than `1e-30`.

python_examples/rebound.py
```diff
@@ -127,4 +149,158 @@ def step():
...
+    o.P = 2.*numpy.pi*numpy.sqrt( o.a*o.a*o.a/mu )  # period
+    o.inc = numpy.arccos( h2/o.h )              # inclination (wrt xy-plane)
+    # if pi/2<i<pi it's retrograde
+    n0 = -h1                                    # node vector
+    n1 =  h0                                    # in xy plane => no z component
+    n = numpy.sqrt( n0*n0 + n1*n1 )
+    er = p.x*e0 + p.y*e1 + p.z*e2
+    if (n<=1.e-30 or o.inc<=1.e-30):            # we are in the xy plane
+        o.Omega=0.
+        if (o.e <= 1.e-10):                     # omega not defined for circular orbit
```
hannorein: Not sure about this one. But `1e-10` seems odd. Either something closer to `1e-16` or `1e-308`.

python_examples/rebound.py
hannorein: (I know that these values come from the c version of rebound, but this might be a good opportunity to think about why they are what they are)

python_examples/rebound.py
dsspiegel: Agreed (math is always available, numpy not necessarily).

python_examples/rebound.py
dsspiegel: I think it would be better to make it forwards compatible, as follows:

    try:
        range = xrange  # this means python 2.x
    except NameError:
        pass # this means python >2.x (swallow exception)

python_examples/rebound.py
dsspiegel: This is super minor, but if any comparisons to None remain, it's more 'pythonic' to write

    if blah is None:

etc.

Also, this is super minor, but please don't put more than a single command on a line.  For instance, if this bit were to remain (although I agree with Hanno it shouldn't) it should be

    if e is None:
        e = 0.
        print('e not passed, setting to 0... (circular)')

If there are other cases where you have more than one command on a line, please change them.

python_examples/rebound.py
dsspiegel: I agree with Hanno, and I think we should name the magic number.  Something like

    SMALL = 1.e-300

at the top of the file, and then

    if star.m <= SMALL

etc.

dsspiegel:
Just wanted to add that this generally looks good, and we're just being picky because now is the right time to do so.  Thanks for the work.

dtamayo:
Hi Dave,

Thanks so much for your quick feedback. I don't have very much coding experience, so I actually find the minor comments quite useful in terms of style etc. So please keep those coming. 

One thing I was curious about is that I  felt like for a new user who may not know much dynamics, it might be better to see warnings and see what variables they're not aware of (for them to look up) than to silently set variables to zero that later they realize make a difference. Since it's an initialization routine I didn't feel like performance was very important, though I agree the code looks a bit messy. Maybe an other option would be to have an optional silent flag you can pass to suppress warnings? Of course, maybe the novice end user is not a likely target we should be thinking about much for this!  It just seems like with a python interface, professors could have students try it for homework sets or things like that. Either way, I'm happy to do it however you see best. Getting your thoughts will help me write future code. 

As for the hard-coded constants in p2orbit, they were taken from the C code, though I agree it's problematic here for dimensional variables (like h). I find from testing that round-trip conversion is accurate to a relative error of ~1e-12. Maybe that's what I should use for all?

dsspiegel:
I think the right way to handle this is to write an informative docstring, so that someone who wants to know the function's behavior can type, "functionname?" and see the information.

My preferred style for docstrings is the following:

    """
    Does blah. (if it's not a void function, this would say 'Returns a list' or whatever.
                if it's an init routine, it would say, 'Initializes a blahblahb struct'
                or whatever.)
    Note that eccentricity (e) and mass (m) are set by default to 0 but can be set to any
    sensible value by setting the appropriate keyword argument.

    Usage
    -----
    returnvar = functionname(arg1, arg2, e=0., m=0., **kwargs)

    Parameters
    ----------
    arg1 : float
           left boundary of the truncated region (that's just an example)
    arg2 : array_like
           a vector of means of gaussian pdfs (again, just an example)
    e : float
        the initial eccentricity, default is 0.
    m : float
        the initial mass (or just 'the mass'), default is 0.

    Returns
    -------
    A Boolean that indicates whether the struct was set appropriately. (Again, just an example)
    """

Obviously, the stuff in parentheses is my own explanation, not stuff that would actually go in the docstring.

hannorein:
My main reason to not have these warning messages is that suppose you want to output the positions every orbit and pipe everything to a text file to post-process. With the warning messages you'd need to write a script that removes them from the output. It's a mess...

dtamayo:
Hi Hanno and Dave,

Thanks a lot. All of this, and especially the reasoning, is really helpful. I appreciate the time and quick replies,

Dan

dtamayo:
I've incorporated all your comments (I think).  I've also refactored init_planet in a way that I think is more useful/general.  Since it returns a particle, I thought it made more sense to call it kepler_particle.  Amusingly, when thinking about what order the parameters should be passed in (so that you can leave out later parameters if you want, e.g., a planar orbit) I ended up passing them in canonically conjugate pairs, which probably isn't a coincidence.  For p2orbit, where possible (for angular momentum), I made the comparisons to check for small values to be dimensionless, which I think is a better way to do it.

p2orbit fails in a lot of edge cases, you can see at least one commented out in test_rebound.py, and that file makes it easy to add in your own test cases to check whether they work or not.  This is going to be a pain and take some thinking, but I don't think it's a huge deal--I think most of the time this is only an issue at initialization (where you'd pass f=pi etc.), so in your output file you would see that the first line doesn't match your input, but everything after that should be right.

hannorein:
Thanks a lot @dtamayo! This is really good. I've added the convenience wrapper for the new kepler_particle() function to the particles_add() function. Now one can use a syntax like:

    rebound.particle_add(m=1e-3, a=1.)

https://github.com/hannorein/rebound/blob/master/python_examples/orbital_elements/problem.py

PR #32, opened 2015-01-23
fixed python 2/3 compatibility issue and fixed some edge cases in p2orbi...

dtamayo:
...t

I realized the range = xrange fix didn't work right, and ran into an edge case for the eccentric anomaly that caused me to crash, so I've made minor fixes to p2orbit.

### DIFF
--- a/python_examples/rebound.py
+++ b/python_examples/rebound.py
@@ -1,9 +1,10 @@
 from ctypes import *
 import math
+
 try:
-    range = xrange  # this means we're using python 2.x
+    xrange          # If there's a name error we're using python 3.x
 except NameError:
-    pass            # this means we're using python >2.x
+    xrange = range  # xrange = range in python 3.x
 
 # Try to load libias15 from the obvioud places it could be in.
 try:
@@ -124,7 +125,7 @@ def particle_get(i):
 def particles_get_array():
     N = get_N() 
     particles = []
-    for i in range(0,N):
+    for i in xrange(0,N):
         particles.append(particle_get(i))
     return particles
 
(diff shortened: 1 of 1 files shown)


hannorein:
Thanks. I merged it. Just made a few stylistic changes in d9211acca

dtamayo:
Great,

Dan

dsspiegel:
I added this comment elsewhere, but I'll add it here too — What was the problem with the range/xrange solution I suggested?

dtamayo:
Hi Dave,

I basically ended up doing what you suggested.  In the original code everything is written with xrange.  When I copy-pasted your solution, I would get a NameError in python 3.x, and the name wouldn't be changed.  I could have changed the xranges in the code to ranges and use what you wrote, but instead I changed the try to check if xrange is defined, and if not change xrange to range.  I believe this achieves the same goal you suggested of making it forwards rather than backwards compatible,

Dan

dsspiegel:
I see.  Yes, my suggestion, though I didn't spell it out, was to switch and use range for everything.  Python 3.x's range is the same as Python 2.x's xrange (they switched the name of the xrange function to range, and they got rid of the range function).

It seems to me cleaner to switch and use the newer name ('range') for everything, instead of encouraging someone who's using Python 2.x to use a deprecated name 'xrange'.  So, my suggestion is to use the solution I originally suggested and to switch usage of xrange to range.

dtamayo:
OK, 

I changed everything in the python directory to conform to this, but for some reason my terminal hangs when I try to push to my repository.  I'll need to troubleshoot this later,

Dan

dtamayo:
OK, I changed things as Dave suggested.  For consistency, we should from now on always use range (instead of xrange) and if rebound.py detects python 2.x it will change all the ranges to xranges.  Also, we should write all print statements in parentheses, e.g. print(x) not print x.  The former works in both 2.x and 3.x, the latter doesn't work in 3.x,

Dan

PR #33, opened 2015-01-24
changed prints and xranges to python 3.x version. Still works in python ...

dtamayo:
...2.x

OK, changed things as Dave suggested.  For consistency, we should from now on always use range (instead of xrange) and if rebound.py detects python 2.x it will change all the ranges to xranges, and we should write all print statements in parentheses, e.g. print(x) not print x.  The former works in both 2.x and 3.x, the latter doesn't work in 3.x.

### DIFF
--- a/python_examples/megno/problem.py
+++ b/python_examples/megno/problem.py
@@ -32,7 +32,7 @@ def setup_planet(com, mass, period, M, omega, eccentricity):
     if eccentricity>0.8:
         E = math.pi
     F = E - eccentricity*math.sin(M) - M
-    for i in xrange(100):
+    for i in range(100):
         E = E - F/(1.0-eccentricity*math.cos(E))
         F = E - eccentricity*math.sin(E) - M
         if math.fabs(F)<1e-16:
--- a/python_examples/dragforce/problem.py
+++ b/python_examples/dragforce/problem.py
@@ -33,4 +33,4 @@ rebound.integrate(100.)
 
 # Output something at the end (the planet will be at ~0.1 AU)
 for i in range(N):
-    print particles[i].x, particles[i].y, particles[i].z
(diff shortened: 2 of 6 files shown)


python_examples/megno/problem.py
```diff
@@ -32,7 +32,7 @@ def setup_planet(com, mass, period, M, omega, eccentricity):
     if eccentricity>0.8:
         E = math.pi
     F = E - eccentricity*math.sin(M) - M
-    for i in xrange(100):
+    for i in range(100):
```
hannorein: Note that this works in python 2.0 -- but it will use the "bad" `range` function. If you overwrite `range` to be `xrange` in the  module, it will not change the namespace in the `problem.py` file.

dsspiegel: Right.  The only thing that wouldn't work is something like `range(5) + range(3)`, which would give a TypeError in Python 3.x but wouldn't in 2.x with the ordinary meaning of range().  So, we in principle might break some backwards compatibility, but only for things that people probably won't do anyway.

hannorein: Well, `range` is also much slower for large N compared to `xrange` (as `range` initializes the entire array first and the iterates over it). But that might not matter for most people (it certainly doesn't in the above line).

dsspiegel: Yeah, but that's unrelated to breaking backward compatibility, right?

Incidentally, and interestingly, `range()` is slower than `xrange()` for initializing arrays but slightly faster for indexing.

hannorein: Well, speed was the reason for using `xrange` in the first place...

dtamayo:
Didn't realize it wouldn't change the problem.py namespace.  Thanks for the clarification,

Dan

hannorein:
You'd have to do something like `from rebound import *`...

PR #34, opened 2015-03-23
fourth order solver

dtamayo:


dtamayo:
Let me know if you're interested in messing around with even higher order.  It's simple to implement, since the higher order derivatives alternate between s2 and s3, multiplied by different powers of beta, and the higher order dXs simply extend the Taylor series in the denominator with the previous value of dX,

Dan

hannorein:
Looks like the higher order one is significantly slower than Euler...

Hanno

dtamayo:
Than Newton?  Interesting, for what kind of orbits and timestep? 

At least at low eccentricities, if I use the initial guess you have of and a 4th order method, I get about 20-30 % faster than Newton's.  Of course that was in Python--maybe it means that in python the X calculations were more expensive in a relative sense as compared to C…

hannorein:
I'm looking at relatively low eccentricities and a timestep ~1% of the period. Thus, I only need ~2-3 iterations with Newton. It's might be hard to get a higher order solver significantly speed this up (we'd need to make sure we only use 1 iteration instead of 2).

Note that I also optimized the C code quite a bit over the weekend...

Hanno

dtamayo:
Right, your changes make a big difference!  

At low eccentricities and dt = 1% of the period, the quartic solver should be converging in one iteration (but then has to do the next to break out).  I could come up with a limit below which we could have it just do a single iteration with the quartic solver, but we can't avoid the second iteration completely, since we need to re-evaluate the G's with the new X, which presumably is the expensive step in the iteration.

It's probably nice that Newton's does best, since it's likely more stable when the derivatives approach 0 (as you've mentioned a couple times), and also because we don't have to come up with as clever an initial guess (which is what makes the higher order solvers better).  This should result in fewer if statements for different regimes, which is a lot cleaner!

Dan

hannorein:
We're still hardly doing better than WH... It depends a bit how it's measured (i.e. WH used to allocate and free its eta array twice per timestep - I changed that to make it a fairer comparison to ours).

Hanno

dtamayo:
Strange,

I need to look to see how they do their calculations.  I've gone through the Stumpff paper in detail hoping it would have some good ideas for the initial guess, but he doesn't take that section very far unfortunately.  It has helped me rewrite the equation in a dimensionless form that makes it clear what the 3 relevant parameters are, and I have a better intuition for what they physically mean.

I'm working now on coming up with a simple initial guess that matches the known solution in the limits of circular and parabolic orbits.

It occurred to me though that in the typical case of low-e orbits, extrapolating from the previous couple iterations may be a much faster solution.  What do you think?

Dan

dtamayo:
Or what about keeping a library of the values over the last orbit and interpolating?  This may work for high e's? 

Dan

hannorein:
WH uses the same guess that you wrote on my board. And that's the one I'm using now too...

I don't like extrapolating from previous steps. The timestep should be self-sustained. Otherwise it becomes a mess when two particles change position or if a particles gets added/ejected...

Hanno

dtamayo:
Ah, OK,

Dan

dtamayo:
Well, just to play devil's advocate, if particles change position or particles get ejected, WH doesn't work anyway…

Dan

PR #36, opened 2015-04-01
Refactor

dtamayo:
These are the changes we talked about to create a separate python_modules folder in rebound/ and separate library folders for librebound and other optional libraries (like libmigration (not included in this commit)) in rebound/shared.  I also split librebound.c into a separate .h file (function prototypes not included yet--need to test with libmigration), and now the python modules and libraries are found a bit more cleanly.

This involves quite a few changes--I settled all conflicts in unaffected files in favor of your versions, but I may have inadvertently added some of my crap in cases where there was no conflict.  Let me know if I can fix this up.


hannorein:
This is just too messy. I will merge the files manually.

PR #38, opened 2015-04-17
modified README for anaconda installation

dtamayo:
I'm pretty sure (but not positive) that if you were in a random directory and installed rebound with virtualenv, you'd have to do

source path/to/rebound/directory/venv/bin/activate

and that the path can be relative or absolute.  I wrote it as such in the readme, but thought I'd double-check with you.

We should probably also remember to add a paragraph for pip install -e . once mikkola is merged into master.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -65,13 +65,18 @@ In short: If you simply want to integrate a few particle system such as a planet
 
 **Python and librebound**
 
-To access REBOUND from python, simple install it via pip. Go to an empty directory where you want to have rebound installed, then type the following::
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -65,15 +65,20 @@ In short: If you simply want to integrate a few particle system such as a planet
...
-    # installation clean.
+    # installation clean (recommended if you are unsure).
     virtualenv venv && source venv/bin/activate
 
     # Then, simple install rebound using pip
     pip install rebound
 
+If you use the Anaconda distribution, the above likely will not work.  Instead create a conda environment (in your anaconda/envs directory) with::
+
+    conda create -n venv pip && source activate venv
```
hannorein: Are you sure this is the right syntax: "source activate venv"?

PR #41, opened 2015-05-27
added Fourier tutorial and added Orbit class in __init__

dtamayo:
I didn't end up using it in the tutorial, but I changed __init__.py to also make rebound.Orbit available (in the same way rebound.Particle is).


hannorein:
That's great! Very detailed! Thank you.

One small comment, the default is `exactFinishTime=1` (note note `exactTime`), so the sampling was uniform in your example. I changed the text a little.

dtamayo:
Whoops! Thanks for catching that,

Dan

rebound/__init__.py
```diff
@@ -15,9 +15,11 @@
 
 from .librebound import ReboundModule
 from .particle import Particle
+from .particle import Orbit
 from .interruptible_pool import InterruptiblePool
 import sys
 module = ReboundModule(__name__)
 module.Particle = Particle
+module.Orbit = Orbit
```
hannorein: BTW, I thought about this, but couldn't come up with a scenario where one would need it. But it's ok to leave it in.

PR #42, opened 2015-06-02
modified installation tutorial

dtamayo:
I think I mentioned that I had problems before getting rebound to work from within jupyter before.  Now that I'm reinstalling everything, I tried to be careful about it, and realized I got things moved around and confused.  Anyway, I changed the tutorial to be more explicit on the anaconda side of things.  I also added a short paragraph at the top about getting python and C.  I figured the people that wouldn't have these installed would probably need the most help, so there are a couple pointers.  I also added a section at the bottom of how to get ipython notebook running to reproduce the tutorials.  Feel free to use only whatever pieces you think are worthwhile.

### DIFF
--- a/python_tutorials/Installation.ipynb
+++ b/python_tutorials/Installation.ipynb
@@ -8,7 +8,9 @@
    "source": [
     "# Installing REBOUND\n",
     "\n",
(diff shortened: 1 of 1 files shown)


dtamayo:
Also, I did test all the commands, using both pip and conda (except for rebound, which must use pip)

PR #43, opened 2015-06-07
Added Julia interface and example

eford:
Most of this is julia code.  
I did add a few accessor/mutators, one function that takes a particle by pointer rather than value, and some compile time warnings when the particle structure has extra members.  
This will make it easier to try out some different samplers, e.g., using the Lora package.


hannorein:
Oh wow! That's fantastic. 

I think it should be perfectly safe to use the direct access to global variables, i.e. without setter/getter functions, just using `cglobal` and `unsafe_load`. 

Let me play with it for a bit before I merge it.

eford:
>
> Oh wow! That's fantastic.
>
> I think it should be perfectly safe to use the direct access to global
> variables, i.e. without setter/getter functions, just using cglobal and
> unsafe_load.
>
Ok.  I wasn't entirely sure how those work.  I just remember that it can be
complicated.

I think global extern vairables are avaliable to external callers in Linux,
but aren't automatically exported in windows.  So we may need to use
setter/setter functions if we wanted it to work with windows/cygwin.  I use
windows laptops, but I did the development/testing for this on linux.

BTW- While poking around in rebound, I noticed a lot of global variables.
Besides being harder to read/maintain/debug, they get in the way of many
compiler optimizations.  If you every have a student who needs to learn the
basics before contributing, you might consider asking them to look into
reducing/eliminating global variables.

Cheers,
E

>
-- 
Eric Ford
Professor of Astronomy & Astrophysics
Center for Exoplanets & Habitable Worlds
Center for Astrostatistics
Institute for CyberScience
Penn State Astrobiology Research Center
Pennsylvania State University

hannorein:
1) I will change some of the function names. We have tried to stick to the convention `verb_noun()`, e.g. `set_integrator()`, at least in the python version.

2) I am aware of the global variables. We've given the compiler a lot of hints how to optimize, especially the kernel parts. It comes down to a matter of taste. I tend to prefer this approach over passing around a large state vector with all the variables to every function call. But it might not stay like this forever... 

3) I think you would be the first person trying to run it on Windows! I think you're right about the global variable issue there. Windows has a different approach to dynamic libraries (and I think it only loads it once into memory, no matter how many programs are using it, or something like that).

hannorein:
It's been a few years, but I'll close this for now. I think it's just not realistic that I will find the time to maintain a C, python, and Julia interface at the same time.

PR #47, opened 2015-06-12
Sync

dtamayo:
I'm pretty happy with this.  Let me know whether you think this is close to what we want, and I'll finalize it / update the tutorials, and add a new tutorial for modifying WHFast incorporating what's already at the bottom of WHFast.ipynb.  Summary:

The defaults are:
recalculate_jacobi_every_timestep = 1
synchronize_every_timestep = 1
corrector = 0
particles_modified = 1

Note that this wasn't the default before (for integrate, the equivalent of synchronize_every_timestep was 0, which was why my code was running so slow (before i set default corrector to 0 from 11).  So if someone was, say, modifying orbital elements each timestep, they would have broken it).

I think the logic is now relatively transparent (and not much different from what you had).  If you change recalculate_jacobi_every_timestep to 0, you can still manually recalculate jacobi elements each time by setting particles_modified to 1 (which then resets to 0 until you set it again).  

I also separated the flags from the call to rebound.integrate() so that they're properties like the other rebound variables.  I'm not sure now whether that's clearer or more convenient.  One advantage would be that the setters can be easily used in any custom steppers people choose to make.  Maybe I should keep the properties but add back in the parameters passed to integrate and have integrate set them?


rebound/librebound.py
```diff
@@ -64,6 +64,14 @@ def additional_forces(self,func):
             # C function pointer
             self.clibrebound.set_additional_forces_with_parameters(func)
 
+    @property
+    def post_timestep_modifications(self):
+        return self.fp   # might not be needed
```
hannorein: This should not be `self.fp` (which is the function pointer for `additional_forces`.

rebound/librebound.py
```diff
@@ -64,6 +64,14 @@ def additional_forces(self,func):
...
+
+    @post_timestep_modifications.setter
+    def post_timestep_modifications(self,func):
+        self.clibrebound.set_post_timestep_modifications(func)
```
hannorein: I think you want to add the same logic as for the `additional_forces` setter, i.e. allow python function to be set too. Note that you'll need to keep a references to the function pointer (for `additional_forces` this is `self.fp`).

rebound/librebound.py
```diff
@@ -314,15 +315,48 @@ def save(self, filename):
...
+    @property
+    def synchronize_every_timestep(self):
+        return c_int.in_dll(self.clibrebound, "integrator_whfast_synchronize_every_timestep").value
+
+    @synchronize_every_timestep.setter
+    def synchronize_every_timestep(self, value):
+        c_int.in_dll(self.clibrebound, "integrator_whfast_synchronize_every_timestep").value = value
+
+    @property
+    def particles_modified(self):
```
hannorein: Would it make sense to call `particles_modified` something like `integrator_whfast_recalculate_jacobi_this_timestep`?

rebound/librebound.py
```diff
@@ -314,15 +315,48 @@ def save(self, filename):
...
 
-    def integrate(self, tmax, exactFinishTime=1, keepSynchronized=0, particlesModified=1, maxR=0., minD=0.):
+    def step(self, do_timing = 1):
+        self.clibrebound.rebound_step(c_int(do_timing))
+
+    def integrate(self, tmax, exact_finish_time=1, maxR=0., minD=0.):
         if debug.integrator_package =="REBOUND":
             self.clibrebound.integrate.restype = c_int
-            ret_value = self.clibrebound.integrate(c_double(tmax),c_int(exactFinishTime),c_int(keepSynchronized),c_int(particlesModified),c_double(maxR),c_double(minD))
+            ret_value = self.clibrebound.integrate(c_double(tmax),c_int(exact_finish_time),c_double(maxR),c_double(minD))
```
hannorein: I think the split you made between attributes and function arguments makes much more sense now. Good!

src/integrator_sei.c
```diff
@@ -5,7 +5,7 @@
  * @detail	This file implements the Symplectic Epicycle Integrator 
  * (SEI). The integrator is described in detail in Rein & Tremaine 2011. 
  * It solves epicyclic motion exactly and is therefore exact up to machine
- * precission in the limit of no perturbing forces. When perturbing-forces
+ * precision in the limit of no perturbing forces. When perturbing-forces
```
hannorein: I don't know just how many times I made this typo...

src/integrator_whfast.c
```diff
@@ -6,7 +6,7 @@
  * 
  * @section 	LICENSE
  * Copyright (c) 2015 Hanno Rein, Daniel Tamayo
- *
+ *l
```
hannorein: ? ;-)

src/integrator_whfast.c
```diff
@@ -731,19 +734,17 @@ static void integrator_apply_corrector(double inv){
...
 		integrator_allocated_N = N;
 		p_j = realloc(p_j,sizeof(struct particle)*N);
 		eta = realloc(eta,sizeof(double)*(N-N_megno));
 		etai= realloc(etai,sizeof(double)*(N-N_megno));
-		recalculate_jacobi = 1;		// Recalculate masses/Jacobi coordinates if first timestep or if N changes.
+		integrator_whfast_particles_modified = 1;
 	}
 	// Only recalculate Jacobi coordinates if needed
-	if (recalculate_jacobi){
+	if (integrator_whfast_recalculate_jacobi_every_timestep || integrator_whfast_particles_modified){
```
hannorein: Again, I think it would make sense to rename `integrator_whfast_particles_modified` to `integrator_whfast_recalculate_jacobi_this_timestep`.

src/librebound.c
```diff
@@ -94,76 +101,41 @@ void rebound_step(void){
...
-	particles 	= NULL;
-	problem_additional_forces = NULL;
-	problem_additional_forces_with_parameters = NULL;
-	integrator_reset();
-	struct timeval tim;
-	gettimeofday(&tim, NULL);
-	srand ( tim.tv_usec + getpid());
+	if (problem_post_timestep_modifications){
+		integrator_synchronize();
+		problem_post_timestep_modifications(particles, t, dt, G, N, N_megno);
```
hannorein: I think you could add this in the `main.c` file too.

hannorein:
I agree, this looks pretty good. The only thing is the renaming of `particles_modified`. But let me know if you think it doesn't make sense.

Thanks for sorting this out!

src/librebound.h
```diff
@@ -52,7 +52,7 @@ void rebound_step(void);
  *   2 = Particle distance exceeds maxR
  *   3 = Close encounter closer than minD
  */
-int integrate(double _tmax, int exactFinishTime, int keepSynchronized, int particlesModified, double maxR, double minD);
+int integrate(double _tmax, int exact_finish_time, double maxR, double minD);
```
hannorein: To be consistent, do you want to call this `rebound_integrate`?

rebound/librebound.py
```diff
@@ -137,16 +137,17 @@ def integrator(self):
     @integrator.setter
     def integrator(self,value):
```
dsspiegel: Super minor suggestion — per PEP-8 https://www.python.org/dev/peps/pep-0008/ you should have spaces after commas in arguments lists.

rebound/librebound.py
```diff
@@ -64,6 +64,14 @@ def additional_forces(self,func):
             # C function pointer
             self.clibrebound.set_additional_forces_with_parameters(func)
 
+    @property
+    def post_timestep_modifications(self):
+        return self.fp   # might not be needed
+
+    @post_timestep_modifications.setter
+    def post_timestep_modifications(self,func):
```
dsspiegel: spaces after commas -- change to `post_timestep_modifications(self, func)`

rebound/librebound.py
dsspiegel: spaces after commas in function arguments.  (sorry for the super minor comments! ... ignore them if you want obviously)

src/integrator_whfast.c
```diff
@@ -41,9 +41,12 @@
 #define MAX(a, b) ((a) < (b) ? (b) : (a))
 #define MIN(a, b) ((a) > (b) ? (b) : (a))
 
-unsigned int integrator_whfast_synchronize_manually 		= 0;
-unsigned int integrator_whfast_corrector 			= 11;
-unsigned int integrator_whfast_particles_modified		= 0;
+// the defaults below are chosen to safeguard the user against spurious results, but
+// will be slower and less accurate
+unsigned int integrator_whfast_recalculate_jacobi_every_timestep = 1;
+unsigned int integrator_whfast_synchronize_every_timestep 		= 1;
```
dsspiegel: I think it's best not to use tabs.

dsspiegel:
Hi Dan (cc Hanno),

I was getting these notifications and just took a quick read through.  Didn't add anything besides some minor linting suggestions.  I hope it wasn't unwelcome ... obviously ignore my comments if you'd prefer.  Looks like you two are doing some great work with whfast!

Best,
Dave

dtamayo:
Thanks Hanno,

I agree that your choice makes things more explicit, and with some documentation (that I'll write) there's no need for the more vague particles_modified. Thanks for spotting the other issues. I want to debug things a bit on Monday with Morgan downtown. For some reason things ran fine on my Mac but she said this evening she was having issues with this new branch (or maybe something else was causing the issue) on Linux,

Dan

dtamayo:
Hi Dave,

No problem at all. I like any and all feedback...I've been learning a lot the last few months...thanks for reading through it and please by all means keep it coming. 

I'm sorry you weren't able to make it to Penn State. It would have been great to meet and talk a bit,

Dan

src/librebound.h
dtamayo: Agreed, done

rebound/librebound.py
dtamayo: Woops, that was a bad one!  I made an additional pointer, and renamed fp for consistency.  In the case we use a C function, I can't see a case where you would want the pointer from within python, but I made a stub anyway...in the case where we use a C function I set these variables to the string "C function pointer value currently not accessible from python.  Edit librebound.py" (I do this also now for additional forces, which didn't have fp set before when passed a C function).

I also made it so you can pass a python function to post_timestep_modifications to parallel what's there for additional_forces

rebound/librebound.py
dtamayo: Ah, thought I'd come up with this myself :)  See above

rebound/librebound.py
dtamayo: ok, i won't guarantee I caught all of them, but changed the ones I saw

src/integrator_whfast.c
dtamayo: I like that better, much less vague.  Done

src/integrator_whfast.c
dtamayo: woops

src/integrator_sei.c
dtamayo: obviously not a big deal, I just change things if I see them

src/librebound.c
dtamayo: Forgot to do that, thanks.  Done.

dtamayo:
I still need to appropriately document the header files to reflect the changes, and will write a modifying WHFast python tutorial incorporating the last section of WHFast.ipynb, so wait on the merge.

dtamayo:
OK, I think I added everything that needed adding.  The Advanced WHFast tutorial is a bit long-winded, but I added a quick summary (do you agree with it?) at the top, and it's probably time to move on :)

dtamayo:
One thing to think about is that I commented out the warning to stderr when you recalculate jacobi coordinates and things aren't synchronized (in part 1).  It seems like this could happen if you modify particles (say you increase the mass) and just set recalculate_jacobi_this_timestep = 1.  Since you didn't synchronize explicitly (it does it when it recalculates jacobi coordinates), you get the warning and things run slow.  What do you think?

hannorein:
I think the warning message should definitely be there.

hannorein:
I'll merge the pull request now and add in... Feel free to send another pull request if you want to change more things.

hannorein:
This broke a lot of the C examples. I think I fixed all of them now. I hope I didn't break anything in your python examples...

dtamayo:
OK thanks,

I’ll check the python examples.  Sorry, I did forget to change the C ones,

Dan

dsspiegel:
Hi guys,

I recommend using unit tests and Jenkins to do regular builds and checks
with a convenient dashboard. Partly because these things are actually
useful, and partly because they're valued skills on the job market.

Dave

On Friday, June 19, 2015, Dan Tamayo <notifications@github.com> wrote:

> OK thanks,
>
> I’ll check the python examples. Sorry, I did forget to change the C ones,
>
> Dan
>
> > On Jun 19, 2015, at 2:32 PM, Hanno Rein <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
> >
> > This broke a lot of the C examples. I think I fixed all of them now. I
> hope I didn't break anything in your python examples...
> >
> > —
> > Reply to this email directly or view it on GitHub <
> https://github.com/hannorein/rebound/pull/47#issuecomment-113599374>.
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hannorein/rebound/pull/47#issuecomment-113606527>.
>


-- 
Please excuse brevity. Gmail Mobile

hannorein:
Thanks for volunteering to implement unit tests, Dave! ;-)

hannorein:
On a more serious note, I think using travis would be more than sufficient. It works really well for the Open Exoplanet Catalogue.

PR #49, opened 2015-06-22
added python_tutorials/ipnbdoctest.py for ipythong notebook testing

dtamayo:
So I found this repository online for automated testing of ipython notebooks.  I chose it because it already had a wrapper for Travis.

It took a bit of tinkering to get things up to date working with both python 2 and 3, and the new ipython messaging specifications, but it's working now.

You can see the options at the bottom of the file, but basically you can check a single file by doing

python ipnbdoctest.py WHFast.ipynb

where you pass it the ipynb to test.  You can pass --diff to output the diff from differences in outputs.  I also found a fork that compares the output from different images in the ipython notebook, so I can try to incorporate that if you think it would be useful.  The only weird thing right now is that in python 2, even though it's not explicitly checking the images, it doesn't give an OK for cells that output images.  Need to figure that out.

Anyway, I haven't looked into how Travis works, but hopefully it's easy to incorporate.


hannorein:
Thanks! Looks promising. I won't have time to look into it over the next few days, but hopefully when I get back.

hannorein:
I'll close this one as per yesterday's discussion.

PR #51, opened 2015-07-08
Id

dtamayo:


rebound/particle.py
```diff
@@ -135,6 +136,11 @@ def __init__(self, particle=None, m=None, x=None, y=None, z=None, vx=None, vy=No
             self.vy = vy
             self.vz = vz
 
+        if ID is None:
+            self.ID = rebound.module.lastID
```
hannorein: I don't think it matters, but there might be a confusion given that `lastID` is really `lastID+1`.

src/boundaries_open.c
```diff
@@ -44,7 +44,7 @@ int nghosty = 0;
 int nghostz = 0;
 
 void boundaries_check(void){
-	for (int i=0;i<N;i++){
+	for (int i=N-1;i>=0;i--){ // run through loop backwards so we don't have to recheck same index after removing
```
hannorein: This *could* (I don't know) be significantly slower. I guess it's a tradeoff depending on how many particles escape per step on average in most cases (even with many particles). I would think that number is small, so would probably favour to go through the loop the normal way.

src/particle.c
```diff
@@ -140,6 +142,42 @@ void particles_remove_all(void){
...
 	N_megno 	= 0;
+	lastID 		= 0;
 	free(particles);
 	particles 	= NULL;
 }
+
+int particles_remove(int index, int keepSorted){
+	if (N==1){
+		printf("Last particle removed.  Exiting.\n");
+		exit(0);
```
hannorein: This has been in there for a while. It might be problematic if this happens within librebound. I don't think anything bad would happen if we do not exit here...

dtamayo:
I made the changes we talked about, updated the tutorials, and decided to add a C example also since it's slightly different.  Unfortunately I can't get the preprocessor directive piece to work (it doesn't recognize PARTICLE_IDS = 1 when I set it in the Makefile).  I tried to compare it to MPI examples that need to do the same thing, and can't tell what's different in what I'm doing.  Could you look at it?  Then I can test whether the example does what it's supposed to.

I also added an error if you try to use IDs with MPI, and a warning if you try to remove a particle and MEGNO is implemented (and then it doesn't remove the particle).

hannorein:
I pushed it to my dtamayo-ID branch. You probably want to update the description of the new C example and add it to the travis build test.

dtamayo:
Thanks,

I'll test it out and update tomorrow morning. I also think I forgot to update the Python index.

dtamayo:
I realized the tutorial/example were almost more about options for removing particles than just the IDs, so I changed their names.  Feel free to change them to whatever you think would be most appropriate.

hannorein:
Please update the description in the new problem.c file. I think you manually updated the readme file instead, but there's a script for that.

dtamayo:
Whoops!  Sorry about that.  Done.

PR #53, opened 2015-07-09
updated p2orbit in tools.c to match python

dtamayo:
This is the C version of the p2orbit that I updated for python.  It fixes several edge behaviors, and is scale-free by checking relative errors.

### DIFF
--- a/src/tools.c
+++ b/src/tools.c
@@ -228,20 +228,7 @@ struct particle tools_init_orbit3d(double M, double m, double a, double e, doubl
 	return p;
 }
 
-struct orbit tools_orbit_nan(void){
-	struct orbit o;
-	o.a = NAN;
-	o.r = NAN;
-	o.h = NAN;
-	o.P = NAN;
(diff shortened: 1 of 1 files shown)


src/tools.c
```diff
@@ -228,64 +228,125 @@ struct particle tools_init_orbit3d(double M, double m, double a, double e, doubl
 	return p;
 }
 
-#define TINY 1.0e-12
-struct orbit tools_p2orbit(struct particle p, struct particle star){
+struct orbit tools_orbit_nan(void){
```
hannorein: Instead of a function I would do this:

    const struct orbit orbit_nan = {.a = NAN,...}

hannorein:
Is the idea to remove the python code and call the c function instead? Or just have everything twice?

dtamayo:
It was to have everything twice, but maybe it makes sense for python to call C in case you want to calculate lots of orbits?

PR #58, opened 2015-07-23
Noglobals2

dtamayo:
Updated Units.ipynb and a few other notebooks, and updated step and integrator_synchronize in simulation.py


hannorein:
Thanks! Still lot's to clean up but it's coming together really nicely...

PR #61, opened 2015-07-30
Alice2

dtamayo:
minor fixes


hannorein:
PS: You forgot to change the variable name in the example which is why travis failed...

dtamayo:
Yeah, sorry about that.

hannorein:
No worries, glad it worked as intended. Thanks for the pull request!

dtamayo:
Yeah, that’s really useful.

dtamayo:
I’m trying to track down a weird error where TestParticles.ipynb fails with clang, but not with gcc, saying with clang that there are no positive values so log scale fails…when I run the notebook myself (and I’ve compiled everything with clang), everything looks normal.

Also, it seems you took the Twitter.ipynb out.  Can I take out the travis test for it then?

Dan

dtamayo:
But now it seems like your build on the same branch succeeded?  Hrm…

hannorein:
I noticed that too. The error appears only sometimes... Confusing.

And yes, I'll remove twitter. With the next syntax, I could't get it into 140 characters ;-)

dtamayo:
I guess I’ll leave it for now unless it comes back up.  I don’t think anything I changed should have affected TestParticles.ipynb.

dtamayo:
Hah

PR #64, opened 2015-08-08
Doc

dtamayo:


dtamayo:
I realized halfway through that it may not make sense to have documentation for particle functions that likely should only get called internally (in particle.py).  Feel free to not use those, or maybe we can keep the documentation in the code but not have it incorporated into the website?).

I changed the simulations to simulation.simulation so that one can call rebound.Particle (marginally) more intuitively.  I remembered what made me want to do that.  I found it kind of awkward that sim.calculate_orbits has the indices for the orbits one less than the corresponding indices in the simulation (since there are N-1 planets).  So I was just looping over particles and doing p.calculate_orbit(sim), and that's when I realized that we were passing the C simulation to all these internal functions rather than sim.  Doing this now is still kind of awkward if you want jacobi coordinates since you need the particle to know what its own index is to get the right com, so I think using the original sim.calculate_orbits makes sense how you have it.

Do you think it would make sense to pad the list passed back by calculate_orbits with an extra empty orbits class at the beginning so that orbits[1] corresponds to particles[1]?  I think it would make loops a bit clearer.  Let me know if you think I'm being silly!

dtamayo:
Travis is really helpful

dtamayo:
Also, most of the unit functions are internal, so I only added docstrings to the public API

hannorein:
Thanks a lot! All looks good. 

I think it's fine to document internal functions. You can always write that this is only intended for internal use. It makes it easier to maintain. But no need to document trivial functions... 

In my opinion padding the orbit list is not very nice. 

Meta comment: The smaller the pull request the easier for me to understand what's going on.

dtamayo:
OK, thanks on all counts.  I’ll try to split up pull requests in the future!

Dan

PR #65, opened 2015-08-09
Addxf

dtamayo:
I don't know why, but the old method for connecting the additional forces pointers doesn't work with the way I have implemented reboundxf.  But everything seems to work fine if I also cast to a C function pointer in the same way you had it for python functions (so you don't have to test what kind of function it is).  Anyway, everything seems to work when I test it (and passes travis) but wanted to make sure I wasn't missing something important.

Also, I have things set up now in reboundxf with a __del__ function that frees (in C) all the pointers in the rebxf_params.  If I understood correclty yesterday, I shouldn't free the xf_params pointer in reb_simulation in reb_free_simulation, because then I'll lose the reference to it and later be unable to free the pointers inside.  Did I get that right?

The other thing I tried for a while (this is in the rebound2 branch of my reboundxf repository), but couldn't get to work is that in addition to the decorator allowing, e.g., 

xf.tau_a = [0.,1e5, 1e5]

for all the particles, it would be nice to be able to do

xf.tau_a[2] = 1e6

but as it stands, this just calls the getter for tau_a, and changes the 2 index of THAT list to 1e6 (but not the value in the C pointer).  Definitely don't try to look into this, but I thought I'd mention it in case you had any ideas.

Thanks for all your help with this!


hannorein:
1) Yes, don't free the xf_params pointer in reb_simulation in the reb_free_simulation function.
2) I don't understand why you had to remove the code which allows for the hookup of python function. I assume this breaks `python_examples/dragforce`?
3) I don't know the answer to the getter/setter question.

hannorein:
Regarding 2). I just saw that you removed the functionality of adding a C-function, not a python-function. In that case the drag force example would still work, but your library would be slow. Still don't understand why that would be the case. Might be faster resolved if we talk in person.

dtamayo:
Thanks Hanno.  I benchmarked python against C, and you were right that that implementation was about twice as slow as the C code.

After a lot of searching online, I still can't figure out how to directly assign a C function to the additional_forces function pointer.  The issue is that when I return the C function from reboundxf, it's of type _FuncPtr, which is the default for functions from a ctypes cdll.  It then gives an error that it's expecting a CFUNCTYPE for additional_forces.  Maybe the reason it worked before is that python was not told in the reb_simulation definition that additional_forces was supposed to be a CFUNCTYPE?

In any case, I did find what I think is a good solution.  If in reboundxf I cast the function I want to a void pointer, and return that void pointer's address, I can cast that efficiently to a CFUNCTYPE.  Despite some effort, I still don't know why it's faster, but it's a solution I found online.  It's also really nice, because it means I don't have to cast it from within the python side of reboundxf--this would require that module to know where rebound is to import it and have the reb_simulation definition (since it's the passed parameter to additional_forces).  So I think that's a huge plus.  

The code in reboundxf looks like this:
def modify_elements():
     func_address = cast(clibreboundxf.rebxf_modify_elements, c_void_p).value
     return func_address

I benchmarked this against the C code and for a particular simulation got
12.4 s vs 10.9 over 1e5 orbits, and
117.5 vs 110.6 over 1e6 orbits,

so it seems it asymptotes to being about 5% slower than the C code, which seems quite good.  Let me know what you think.

hannorein:
Hm. I need to think about this for a bit. I don't understand everything.

hannorein:
So I tested what's going on. I think what you did is good. But please keep the python function pointer functionality in there. I don't see a reason as to why we want to remove that.

FYI, here's a small test example of how to use both c and python functions as a callback:

https://github.com/hannorein/rebound/tree/testlib/testlib

dtamayo:
Hi Hanno,

Thanks a lot for looking into it so quickly.  The functionality to use a python function hasn’t changed.  It’s the C function pointer functionality that now looks exactly like the python functionality in simulation.py, so I don’t check for the type that’s being passed.  

The way it’s set up, reboundxf does exactly this from your example:

beat_addr=  ctypes.cast(ctestlib.testlib_beat, ctypes.c_void_p).value

and then simulation.py does 

sim.heartbeat = sim.AFF(beat_addr)

which is exactly what we also need to do when passing python functions to the additional_forces property,

Dan

dtamayo:
That’s why Forces.ipynb still works,

Dan

hannorein:
Oh I see. Hm. So does this function call go via python every time?

dtamayo:
That’s what I can’t figure out directly.  I tried benchmarking the python vs C, and got that the time difference was asymptoting to about 5% slower for python (vs a factor of 2 before for the same example), so I want to say no, but I can’t guarantee it.

hannorein:
I see. Hm. Since you have a benchmark. Can you test this idea?

https://github.com/hannorein/rebound/commit/92e9c1b5715d3d85be7cc0c0284e22e5187e17a7

This really shouldn't go via python because it just sets the address.

dtamayo:
Ah perfect.  Ok I will do that.  I’m leaving for the conference now, but will get that done today.  Thanks for all your help,

Dan

PR #76, opened 2015-08-27
changed void pointer name

dtamayo:


hannorein:
Probably also want to change it in simulation.py

dtamayo:
Thanks, done

PR #80, opened 2015-09-10
fixed simulation.__del__

dtamayo:


hannorein:
Nice and clean!

dtamayo:
I considered breaking it into several commits ;)

hannorein:
;-)

BTW, one related thing that could be fixed in the long term is "copying". Right now one cannot make a deep copy of a simulation. Very, very low priority though...

PR #81, opened 2015-09-21
Dt

dtamayo:
changed dt_last_success in ias15 to use r->dt_last_done


hannorein:
Looking at all the occurrence of `dt_last`, it might be a bit more complicated. I need to think about this.

dtamayo:
OK.  I checked for occurrences of dt_last_success, but didn’t think to look into dt_last_done.  Looking now I think I see what you mean...

dtamayo:
I *think* this is a fix, but you might want to look at it more closely.  My logic was:

In rebound.c, the value in dt_last_done is only used (i.e., the only place it shows up on the right hand side of an equal sign) is after the integration is done and you want to reset it to its original value.  So I think it's safe to just store the original value in a local variable.

dt_last_done is updated to hold the value of dt, but I think inconsistently.  For example, with the previous code, if you run this

import rebound
sim = rebound.Simulation()
sim.add(m=1.)
sim.add(a=1.)
sim.dt = 0.1
sim.integrate(0.05, exact_finish_time=1)
print(sim.dt_last_done)

you'll get 0.1, rather than 0.05.  

I thought it would make sense to update dt_last_done right after a call to reb_step, though the tricky part is that ias15 can choose a different timestep.  So I updated dt_last_done before the call to reb_step, and then allow ias15 to change it if it wants to.  This fixes the problem above, and it fixes the discrepancies I was getting with ias15 with reboundx.

The only problem with it is that in integrator_ias15.c, it checks whether dt_last_done != 0 to see whether you're on the first step (and won't predict the timestep if it's the first one).  Because I set dt_last_done before the call to reb_step, dt_last_done is always != 0.  Is there another way to tell ias15 that it's the first timestep?

hannorein:
Thanks! That makes all sense. I think there is no really issue with not detecting the first timestep (other than a few extra calculations in the first timestep). But let me stare at it for a little longer...

hannorein:
I think we're not quite there yet. A few points... might be faster to go through this together in person:

- dt_last_done will be finite before calling reb_step for the first time.
- IAS15 needs to know the last dt that it has done successfully to extrapolate the coefficients for the polynomials. I think this is not working right now. dt_last_done will be updated if an IAS15 timestep fails and IAS15 has to reduce dt before try again (in that case it needs the old dt before it reduced it).

hannorein:
Ok. I think this makes sense now!

dtamayo:
I think I checked all the checked all the cases, so as long as this passes Travis I think it should be good.  I also just went with the simple setting last_full_dt to dt at the beginning since it's a local variable the user wouldn't use anyway.  This makes the logic much clearer

PR #82, opened 2015-10-20
Added additional parameters void pointer and sim pointer to particle struct

dtamayo:
I think this works, but wait for Travis


src/rebound.h
```diff
@@ -107,6 +107,8 @@ struct reb_particle {
 	double lastcollision;		///< Last time the particle had a physical collision.
 	struct reb_treecell* c;		///< Pointer to the cell the particle is currently in.
 	int id;				///< Unique id to identify particle.
+	void* ap;			///< Functionality for externally adding additional properties to particles.
+	struct reb_simulation* sim;
```
hannorein: How about a documentation string here?

hannorein:
Looks simple and clean!

dtamayo:
Sorry, I had that in there before and forgot to put back in.  I also added something for _sim in Particle...I thought it was better to document that it shouldn't be used, rather than letting people wonder and try it?

hannorein:
Agreed!

PR #84, opened 2015-10-21
Fix double-counting collisions in collision.c

joshuawallace:


hannorein:
I think there might be an issue when using periodic or sheer-periodic boundary conditions. Let me think about it for a bit longer...

hannorein:
I'm going to close this because this will not work for periodic boundary conditions. The current implementation might not be as fast as it could be, but it is very general.

PR #85, opened 2015-10-21
P orb

dtamayo:
New decorators work for orbital elements:
->particles[0].any_element returns 0 and prints a warning saying it's not implemented
->elements always returned in jacobi coordinates

I also removed the last spots where simulation.py repeats C code and have it now calling C in sim.calculate_com and sim.calculate_orbits.  I also added an option in calculate_orbits to use barycentric coordinates (in addition to jacobi and heliocentric).

In C, I added 2 functions in tools.c:
reb_get_particle_index takes a particle pointer and returns its corresponding index. 
reb_get_jacobi_com takes a particle pointer and returns its jacobi com


rebound/particle.py
```diff
@@ -254,3 +254,65 @@ def calculate_orbit(self, simulation, primary):
...
 
         return o
+
+    def orb_el(get_elem):
+        @property
+        def new_elem(self):
+            clibrebound.reb_get_particle_index.restype = c_int
+            index = clibrebound.reb_get_particle_index(byref(self)) # first check this isn't particles[0]
+            if index == 0:
+                print("Orbital elements for particle[0] not implemented.")
```
hannorein: This should probably be an exception rather than a print statement.

src/tools.c
```diff
@@ -232,7 +255,7 @@ struct reb_particle reb_tools_orbit2d_to_particle(double G, struct reb_particle
 	return reb_tools_orbit_to_particle(G, primary, m, a, e, inc, Omega, omega, f);
 }
 
-static const struct reb_particle reb_particle_nan = {.x = NAN, .y = NAN, .z = NAN, .vx = NAN, .vy = NAN, .vz = NAN, .ax = NAN, .ay = NAN, .az = NAN, .m = NAN, .r = NAN, .lastcollision = NAN, .c = NULL, .id = NAN};
+static const struct reb_particle reb_particle_nan = {.x = NAN, .y = NAN, .z = NAN, .vx = NAN, .vy = NAN, .vz = NAN, .ax = NAN, .ay = NAN, .az = NAN, .m = NAN, .r = NAN, .lastcollision = NAN, .c = NULL, .id = -1, .ap = NULL, .sim = NULL};
```
hannorein: Is there a reason why id is -1 now (I don't think it matters, just curious).

hannorein:
I love the decorator, by the way.

src/tools.c
dtamayo: I had it as NAN, and was getting a warning about overflow trying to assign NAN to an int

hannorein: Good point.

dtamayo:
Thanks!  Seemed like the perfect tool and I wanted to learn how to write my own.  Certainly saves a lot of text!

Issue #73, opened 2015-08-19
Exact Finish Time


hannorein:
Currently, the exact finish time variable is inconsistent between python and c. Once again. Certainly my fault. 

The question is how to choose the best default behaviour. Right now c has it set to 0 and python to 1.

Idea?

dtamayo:
I have mixed feelings.  On the one hand you want to protect the user in cases like using WHFast where you don’t want exact times.  On the other, it can take a long time and be quite frustrating to figure out that the program is not by default giving you the exact times you ask for.  I think you convinced me of the latter point of view last time.  Maybe one possibility would be to add exact_time = 0 behavior in whfast_safe_mode?

dtamayo:
The more I've tried different things, the more you've convinced me that the right behavior should be exact_finish_time = 1.  If you ask to integrate to t=time, that's what you should get.  I think it should just show up prominently in the WHFast documentation that you probably want to set exact_finish_time = 0.  What do you think?

hannorein:
I tend to agree. We'd need to go through all the tutorials and examples to make sure it's consistent.

dtamayo:
OK.  That would also give me a reason to go through the ipython examples and change the calculate_orbits to use the new particles[i].element format.  Unless you have an objection to that, do you want to split it up—I’ll do the ipython examples and you can do the C ones?

Also, the reason we had calculate_orbits in simulation.py was because you couldn’t do jacobi coordinates from within particle.calculate_orbit, since it needed to know about the other particles.  Now we don’t have that limitation, and I think it makes sense to compartmentalize that functionality within the particle class.  Then if you had a thousand planetesimals and a planet, you could calculate the elements of the planet all at once, but without doing the entire simulation.  Would you agree with that?  If so, I can remove sim.calculate_orbits and update all the ipython examples at once.

hannorein:
Hm. I need to think about the second part.

dtamayo:
OK, I could be wrong about that.  It just seems to me like if the state of the particle is in the Particle class, the functionality to convert that state to orbital elements should also be in there.  Right now we have a function that the user isn’t supposed to call (particle.calculate_orbit) and one they should (sim.calculate_orbits()).  This seems like it would be more flexible by letting the user use it like calculate orbits:

orbits = [sim.particles[i].calculate_orbit() for i in range(1,sim.N)]

or just get an individual orbit = sim.particles[3].calculate_orbit().

[...]

PR #88, opened 2015-10-25
Dt exact finish time

dtamayo:
I updated all the ipython notebooks that needed it, and changed to particles[i].orb_el where I thought it made it simpler (this left several sim.calculate_orbits in cases where it was done as part of a loop).

I also (after all) moved my orb_el decorator into Particle.calculate_orbit, which I think is a more logical place for that logic to be.  Now all the orb_els are properties that call calculate_orbit.  For convenience, I also added a last property particles[i].orbit that is equivalent to calling calculate_orbit and assuming Jacobi coordinates.  I thought, particularly since we are discouraging the user from using Particle.calculate_orbit, that it completed the set of orbital element properties.

Sorry, I should have split the particle.py changes into a separate commit--it might get lost in the diff among all the ipynb changes.


hannorein:
Closes #73

hannorein:
One way to keep the ipynb noise down a bit is to reset the kernels, and then run all the cells from top to bottom, before the commit.

PR #89, opened 2015-10-25
changed id default to 0 and updated ipynbs to match

dtamayo:
I thought it made sense to change the default particle ID to 0 (was -1), since in C when you do reb_particle p = {0}, everything defaults to 0, including the ID, so it would be nice to be consistent with both.  I updated particle.py, the documentation and relevant ipynbs (didn't overlap with the ones I modified for exact_finish_time so hopefully shouldn't be a problem for merge).

### DIFF
--- a/rebound/particle.py
+++ b/rebound/particle.py
@@ -92,7 +92,7 @@ class Particle(Structure):
             True longitude              (Default: 0)
         r           : float       
             Particle radius (only used for collisional simulations)
-        id          : int         
+        id          : int               (Default: 0)
             Particle ID (arbitrary, specified by the user)
         date        : string      
             For consistency with adding particles through horizons.  Not used here.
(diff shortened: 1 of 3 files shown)


rebound/particle.py
```diff
@@ -119,7 +119,7 @@ def __init__(self, particle=None, m=None, x=None, y=None, z=None, vx=None, vy=No
         orbi = [primary,a,e,inc,Omega,omega,pomega,f,M,l,theta]
         
         self.m = 0. if m is None else m
-        self.id = -1 if id is None else id
```
hannorein: By the way, I personally find this syntax a little hard to read. The following seems more straight forward:

```python
if m is None:
    self.m = 0.
else:
    self.m = m
```

dtamayo: OK, that sounds good

PR #90, opened 2015-11-02
set sim pointer to particles when loading sim from binary

dtamayo:
When you loaded a simulation from a binary, the particle pointers to the simulation were not set, so since all the calculate_orbit routines rely on these, none of these worked, which also messed up OrbitPlot etc.


hannorein:
Makes sense. Thanks

PR #91, opened 2015-11-04
explanation for offset of few days in 67p ipynb

dtamayo:
I just updated the note at the bottom of Churyumov-Gerasimenko to not say we calculate a date in february, and to say it's because of jets from the comet.

I tried restarting the kernel and running all the cells, but still get a huge diff.  I wonder whether this is because we have different setups (e.g. Python2 vs Python3).


hannorein:
Yeah, the long diffs are annoying. And yes, it's probably due to the different python versions. But I think it's actually good that we are using different versions. This way we noticed if the code doesn't work on one of them...

dsspiegel:
@hannorein @dtamayo,
To avoid huge diffs, you can write your notebooks in markdown and have them automatically converted to ipynb: https://pypi.python.org/pypi/notedown/1.0.3
Then, don't pay attention to diffs in the .ipynb files — just look at diffs in the (human-generated) .md files.

dtamayo:
Interesting!  Thanks Dave

hannorein:
But we'd loose the images (or need to do something fancy).

dsspiegel:
It's true.  This fix is for notebooks that will contain code (and
description) that you want the user to execute.  If you want the notebook
to be pre-populated with figures, it's a little harder to make it work
entirely in markdown.  (You could upload the images to a server that's
web-accessible and show the images via urls embedded in markdown if you
want...)

On Wed, Nov 4, 2015 at 10:33 AM, Hanno Rein <notifications@github.com>
wrote:

> But we'd loose the images (or need to do something fancy).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hannorein/rebound/pull/91#issuecomment-153822239>.
>



-- 
Dave Spiegel, Ph.D.

PR #92, opened 2015-11-04
changed if x is None else x blocks

dtamayo:
I thought in the ID pull request you had changed the lines of the form
self.m = 0. if m is None else m

to
if m is None:
  self.m = 0.
...

but just realized you just wanted to run that by me first.  I just changed those lines.


hannorein:
It was more of a random comment. But thanks! I think this one is more readable.

PR #93, opened 2015-11-05
added _extras_ref to simulation

dtamayo:
I ran tests and confirmed that we did in fact have a memory leak with the self.ref line in __init__, and now with the extras_ref, writing a create_sim function that only returns sim keeps reboundx alive

### DIFF
--- a/rebound/simulation.py
+++ b/rebound/simulation.py
@@ -186,7 +186,6 @@ class Simulation(Structure):
     """
     def __init__(self):
         clibrebound.reb_init_simulation(byref(self))
-        self.ref = byref(self)
 
     @classmethod
     def from_file(cls, filename):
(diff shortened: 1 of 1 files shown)


hannorein:
I think we never see `_b_needsfree_`. This doesn't seem right?

dtamayo:
What do you mean?  _b_needsfree is something from ctypes

hannorein:
oh, I forgot that.

dtamayo:
This came in because sim.particles[i]._sim.contents.G was creating a Simulation instance, immediately going out of scope and freeing all the C memory, so I found this flag in ctypes.  We could add in our own in __init__, which I think would serve the same purpose of only freeing the memory when it’s an instance that got created the “normal” way.  I’m not sure which is better.

Build seems to have errored timing out in horizons, I thought you took that out?

hannorein:
Hm. Build seems to have passed according to the above. I don't think I took all of the horizon stuff out.

dtamayo:
Woops!  I got an e-mail saying it had failed, but that was from me pushing to my own fork, where I haven't taken out Horizons.ipynb.  My fault.

PR #96, opened 2015-11-19
moved class variables to instance variables

dtamayo:
This solves the case you sent me.  Probably worth letting it run through travis to make sure it didn't break something.  I think it should be fine though.


dtamayo:
I think this is a fix.  We now don't have any class variables or any variables set in __init__.  And now the getter and setter for units check for whether self has a _units attribute, and if not, sets it to a dictionary of Nones.  Not a permanent fix, but fixes the cases we talked about while I think of something better.

hannorein:
That's a good solution!

dsspiegel:
This is totally unimportant, but I'm curious why `_fields_` is a list of tuples.  It looks like it's a collection of key-value pairs, for which a dictionary seems like a natural data structure (does the order of the tuples in the list ever matter? ... even if so, maybe an OrderedDict would work?).

hannorein:
Note that this is a ctypes variable (not outs). Since the order matters, I suspect they want to be backwards compatible. OrderedDict was only introduces in 2.7.

dsspiegel:
Ahh, then I have a more substantive comment.  It would be a really good idea to change the `from ctypes import *` and `from .particle import *` lines.  It makes the code impossible to trace (I was wondering what `Structure` was...).  I think it's a really good idea to have the code be entirely self contained in the sense that if you want to find out where a variable, class, function, or method comes from, you can explicitly see the origin in the source code.

hannorein:
I agree. Feel free to send me a pull request ;-)

dsspiegel:
I'd be happy to, but I actually don't know what's used from ctypes and from .particle.  It would be a lot easier for someone familiar with what comes from where to write that down instead of someone else needing to dig for it.  (Note that "someone else" could be the original author in 2 months.)

dtamayo:
I have mixed feelings about this. I was of the same opinion so to give examples from our code at pycon I rewrote the snippets to say ctypes.Structure etc. But I realized it makes some sections of the code extremely wordy where there are 5 or 6 ctypes objects per line. I find it way clunkier to read, and while I agree someone new to the code could be confused, it is after all a wrapper entirely written with ctypes. As long as we don't import * from other modules, it's maybe not that ambiguous.

hannorein:
Stop guys, please.

dtamayo:
Haha

PR #98, opened 2015-12-09
Satellite evolution

r-cloutier:
Here's my pull request attempt for the last ast3020 assignment. The problem file integrates 100 test particle satellites around Jupiter for a short tmax to ensure that the simulation does not take too long. The extra script contains a function used to compute the joviancentric orbital elements of the satellites.

-Ryan

### DIFF
--- a/python_examples/example1
+++ b/python_examples/example1
@@ -1,0 +1,1 @@
+im an example
--- a/python_examples/satellite_evolution/compute_satellite_orbits.py
+++ b/python_examples/satellite_evolution/compute_satellite_orbits.py
@@ -1,0 +1,75 @@
+import rebound
+import numpy as np
+
+def compute_satellite_orbits(self, outname='satellite_orbits.dat'):
+   '''Compute the semimajor axis, eccentricity, and inclination of 
+   each satellite in the simulation. The resulting orbital elements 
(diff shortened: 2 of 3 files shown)


hannorein:
Hi Ryan, 

Thanks for the pull request. It's a good example! I'll give you a few very minor comments. Maybe they become useful at some point. 

- You don't need to calculate the orbital parameters yourself. You can do 
           
        print sim.particles[2].a 
        
 or, because you are interested in the orbital parameter of the moon with respect to the planet:
  
       ps = sim.particles
       print ps[2].calculate_orbits(primary=ps[1]).a

- Because it's an example, I would probably not store the orbital parameters in a file, but just keep them in memory until plotting.

- Matplotlib tries to open an X window upon import. For people who run the code remotely or don't have an X server running/configure, this would be a problem. An easy fix is to do
        
        import matplotlib; matplotlib.use("pdf")
        import matplotlib.pyplot as plt

 instead of

       import matplotlib.pyplot as plt

- Instead of calculating the value for G explicitly
         
        k = 0.01720209895
        sim.G = k*k           # Gravitational constant [M=Msolar, L=AU, T=days]

 you could just use
  
       sim.units = ["AU", "MSun", "day"]

--
Hanno

PR #99, opened 2015-12-10
Simulating the Kepler-34 Circumbinary System in Rebound; Exploring Add'l Planets

alphaparrot:
This example is an iPython notebook, intended to be accessible even to beginning programmers and newcomers to dynamics or rebound, modeling the Kepler-34 binary system. Kepler-34 consists of two sun-like stars, each with roughly 1 solar mass,  in eccentric (e=0.5) orbits with a semimajor axis of 0.2 AU. A gas giant is very nearly coplanar and in a slightly eccentric orbit, with a semimajor axis of roughly 1 AU, and a period of 288 days. Additional planets further out are quite possible, but additional planets interior to K34(AB)b are unlikely.


hannorein:
Hi Adiv, 

Thanks for the pull request. It's a super detailed example! I'll give you a few very minor comments/questions. 

- You say you are not happy with REBOUND not being able to initialize barycentric orbits. Then calculate `rp` and `ra`. But you then never use these variables again. I'm a bit confused. I'd love to know a better way to initialize binary orbits if the current one is not making sense to you!

- All the examples are automatically executed whenever someone makes a commit or creates a pull request. This is to check is any changes in the code break one of the examples. For that reason they should all run in a few seconds each so that all of them together take less than 30 minutes (which is the upper time limit on Travis CI). 

- I'm not sure if this is important, but I'd be careful using the Star Wars image. I don't want to get a lawsuit from Disney. 

- As it turns out, I worked with an undergraduate student on this system this year. Adding additional planets is exactly what I asked the undergraduate student to do, so I'm really impressed with you coming up with this yourself.

--
Hanno

alphaparrot:
Hi Hanno,

Thanks for the feedback! I'm not sure about the platform you're running 
on or how it long it took you, but from what I recall, running them all 
together on my own laptop definitely takes less than 30 minutes--the 
longest integrations are maybe a few minutes at the most (just long 
enough to make a cup of tea), and there aren't enough to push it over 
the 30-minute limit, I don't think. I could be wrong though.

The main issue I had with initializing binary orbits was that REBOUND 
doesn't seem to calculate an orbit for the first object in 
simulation.particles. That's treated initially as a central object. So 
initializing in a barycentric frame by putting the barycenter at 0,0 and 
starting the stars off at their appropriate positions in x,y barycentric 
space with the right velocities for whatever reason always resulted in 
one or both being on escape trajectories (the case for both being if the 
first item in particles[0] was a test particle of negligible mass and 
zero velocity at 0,0), despite having the correct velocities for their 
orbits (I spent about 7 hours going over the math and deriving and 
re-deriving the vis viva equation in barycentric coordinates to be 
sure). For whatever reason, that problem goes away if you start off in 
the frame of reference of one of the stars, and let that be the primary, 
and then move the simulation to a barycentric frame. My guess, looking 
through REBOUND's source code, is that the semimajor axis calculation 
that REBOUND does may not be sufficiently general if the user manually 
creates a barycentric system (the calculation is just coming from the 
vis viva equation, and without sufficient barycentric adjustments to the 
supposed central mass, the vis viva equation is a bit more complicated), 
but I also checked that math, and REBOUND seems to get the same answer I 
do with my own calculations, so I'm really not sure where the hyperbolic 
orbits come from. But given that most binary star systems are anyway 
presented in the literature in terms of their mutual separation rather 
than distance from barycenter, the workaround I used is not a terrible 
inconvenience.

Adiv

hannorein:
RE 30-minutes: That's combined for all (50 or so) examples and tests. Thus each one needs to run in a few seconds, otherwise we'll have to wait for an entire day.

RE binary: I'm not sure I fully understand. You can give REBOUND any set of cartesian position and velocity coordinates and it will get it right. If you initialize a particle with orbital parameters, it needs to know around which other particle the new particle orbits. One could create a wrapper function that initializes two particles at once, that would do basically what you've done. Maybe we can talk about it in person some day. I'd like to understand what the typical use case would look like and how to make it as intuitive as possible.

alphaparrot:
Re 30 minutes: Ah, I understand. The run-time for the example can be 
drastically reduced by reducing the number of timesteps in each run (but 
going to the same tmax)--the time resolution is artificially higher than 
necessary simply to make the plots look better. Far fewer timesteps 
would make the curves look less smooth, but accuracy shouldn't be 
compromised, since it's IAS15. That would also have implications for the 
plotting commands, so I can make that change if you want. My most recent 
version of the ipython notebook also includes adding a Titan-like moon 
to K34(AB)b, but that was done last weekend after submitting the 
assignment, so hasn't been committed to the github repository (so you 
wouldn't be grading later work). I can keep that in the updated notebook 
if you want, or leave it off so the modification is just for the 
submitted assignment.

Re binary: I found the problem. I was missing a factor of m1/(m1+m2) in 
my velocity calculations. With that added in, the orbits work as 
expected. I've written the derivation and cartesian binary 
implementation up in a separate notebook; if you'd like it as an example 
I can send it to you or submit it as a pull request. In any case, I 
think it represents an atypical use case for someone implementing a 
binary system--it means they've taken the stellar catalog information, 
which provides mutual separations, eccentricity, orbital period, etc, 
from those they derived velocities relative to the barycenter, and 
computed initial conditions in a cartesian barycentric system, rather 
than initializing Rebound with orbital parameters directly from the 
catalog in the frame of one of the two stars (as I did in the 
assignment). The approach I initially tried of initializing Rebound with 
just positions and velocities relative to the barycenter is a far more 
involved and difficult approach than the one I ended up using, which 
requires almost no modifications to the given parameters. But it may be 
worth having the derivation and implementation example on hand.

Adiv

hannorein:
Sounds all goof. If you want, you can commit and then push the changed to github. New commits will get automatically added to this pull request.

zhexingli:
Just curious, where's this example? I don't see it in ipython example folder. I'm planning to run something similar using REBOUND and I'm new to this. It would help greatly if this is available.

hannorein:
I'm embarrassed to say that I never got around to merging it!

https://github.com/alphaparrot/rebound/blob/master/ipython_examples/Kepler-34.ipynb

zhexingli:
Haha, thx for providing the link!

zhexingli:
REBOUND uses G = 1 by default and this mean everything are in units of AU, solar mass, years/2pi. The two stars in this case orbit each other every 28 days = 28/365.25 = 0.076 years. So shouldn't T be 0.076/2pi = 0.0122 instead of 0.076 years. Similarly for the planet b's period below, why the 2pi is not applied?

PR #100, opened 2015-12-11
Docversion

dtamayo:
I'm doing the docs for REBOUNDx and noticed that the version was not getting updated in the REBOUND documentation.  I modified update_version to also update doc/conf.py automatically.


hannorein:
Thanks!

PR #101, opened 2015-12-11
Spacecrafts in Solar System

JuHamp:
The added example calculates trajectories of spacecrafts into the future, determines when they will leave the solar system and creates a plot of the trajectories in the solar system in x-y plane.

### DIFF
--- a/python_examples/spacecrafts/problem.py
+++ b/python_examples/spacecrafts/problem.py
@@ -1,0 +1,61 @@
+#import rebound module
+import rebound
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+#include objects from solar system in simulation
+solar_system_objects = ["Sun", "Mercury", "Venus", "Earth", "Mars", "Jupiter", "Saturn", "Uranus", "Neptune", "-125544", "-98","-24",  "-31", "-32", "-23"]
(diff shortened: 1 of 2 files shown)


hannorein:
Hi Julia, 

Thanks for the pull request. It's a good example! I'll give you a few very minor comments. Maybe they become useful at some point. 

- For it to be useful for a new users, you could add a short paragraph at the top, explaining what you are about to do. 
- You wouldn't want to add temporary files such as the `problem.py~` to git. No need to keep track of that.

- Matplotlib tries to open an X window upon import. For people who run the code remotely or don't have an X server running/configure, this would be a problem. An easy fix is to do
        
        import matplotlib; matplotlib.use("pdf")
        import matplotlib.pyplot as plt

 instead of

       import matplotlib.pyplot as plt

- You're using the WHFast integrator. That works well for bodies orbiting the Sun, but it will not work well for bodies orbiting planets (e.g. the ISS). A specialized symplectic integrator would be needed for that, or, use the allrounder, the IAS15 integrator. 

- All the examples are automatically executed whenever someone makes a commit or creates a pull request. This is to check if any changes in the code break one of the examples. For that reason they should all run in a few seconds each so that all of them together take less than 30 minutes (which is the upper time limit on Travis CI). In your case querying the NASA database takes most of the time...


--
Hanno

PR #102, opened 2015-12-11
python example of restricted three body problem

aobertas:
Added a python example of the restricted circular three body problem. The plotting file will also show the Jacobian for the initial conditions.


hannorein:
Hi Alysa, 

Thanks for the interesting pull request. I love the contour plot! Some of the orbits look really complicated!

A few minor comments.

- When you type a number as `2` into python version 2, it is interpreted as an integer, if you type as `2.0` it is a floating point number. The difference is `1/2=0` but `1.0/2.0=0.5`. A good practice is to always use the dot in numbers unless you want to explicitly define an integer, e.g. for a loop. I've made this mistake a few times myself and wasted days to find the bug. 

- Note that the other examples are all executed automatically whenever someone makes a commit or creates a pull request. This is to check if any changes in the code break one of the examples. For that reason they are all able to run "head-less", i.e. saving the plot to a file instead of displaying it on the screen. Especially since it runs so quickly.

- Because it's an example, it might make sense to add the plotting routines into the main file. Then there is no need to save anything to a temporary file. (PS: I had to change the filename in plot.py to be able to run the code).

- There are couple of things that could be cleaned up. For example, you never use some of the parameters you store in the file. One could also add a bit more explanation as to what the Jacobi constant is and why it is interesting to plot. 

- Your example finishes in a very short time which is great because all examples are always executed when someone changes any part of the code to check if the changes break anything. 

--
Hanno

PR #103, opened 2015-12-11
Resonances of Jupiters moons

lokhorst:
Hi Hanno,

I've created two new examples written in the .ipynb format.  My main submission for the assignment is Resonances_of_Jupiters_moons.ipynb.

I am also submitting Planet_disk_interactions.ipynb, which is another example that I wrote. 
This one is similar to the spreading_disk and restricted_three_body examples, but I've converted it from C to .iypyng format, so I am also sending a pull request for that file.  
I thought since I did that work, it would be worth it to send it to you if you did want to do something like that for the tutorials.  

Cheers,
Deborah


hannorein:
Hi Deborah, 

Thanks for the fantastic pull request. It's super detailed and it was great to see that you were able to use so many different functions of REBOUND. And I like that you made use of the somewhat more complicated tree method! 

A few minor comments.

- Interestingly, when I run the planet disk interaction example, the planet moves away form the star. Not sure why. Maybe just random noise? Or do you think it should always move inwards?

- Note that in an iPython notebook, you can use headings and text, not just comments in the python code. That makes it look a bit nicer.

- The resonances example was missing a line:

      a = np.zeros((Nmoons,Nout))
      
- All the examples are automatically executed whenever someone makes a commit or creates a pull request. This is to check if any changes in the code break one of the examples. For that reason they should all run in a few seconds each so that all of them together take less than 30 minutes (which is the upper time limit on Travis CI). I'll need to wiggle a few nobs to make yours runs a bit faster. 

--
Hanno

lokhorst:
Hi Hanno,

Thanks so much!

I think that the planet moving outwards is probably random noise.  I thought it might be nice to do a more statistical approach, but that would take a while to run.
Sorry that line was missing!! I think I decided to take it out last minute because I didn’t end up using it, but then forgot to take it out of the loop.  That was a silly mistake.
Also sorry they are too slow — were they really half an hour long? I only had to run them for maybe a minute.  Sorry about that.
 
Thanks again,
Deborah

hannorein:
No, yours was only a few minutes! But there are about 50 examples and tests in total. If every one of those takes a few minutes, it'll be a few hours in total...

PR #104, opened 2015-12-11
Added spacecrafts as commit

arachkov:
Third attempt on pull request.

### DIFF
--- a/python_examples/spacecrafts/problem.py
+++ b/python_examples/spacecrafts/problem.py
@@ -1,0 +1,110 @@
+import matplotlib; matplotlib.use("pdf")
+import matplotlib.pyplot as plt
+import rebound
+import os.path
+import os
+import numpy as np
+
+filename = "cache.bin"
+
+# Solar system objects: Sun and the eight planets + Pluto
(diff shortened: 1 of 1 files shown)


hannorein:
Hi Alex, 

Thanks for the pull request. I'll give you a few very minor comments, similar to what I wrote for #101. Maybe they become useful at some point. 

- You probably want to set a proper e-mail address in your git config. 
- For it to be useful for a new users, you could add a short paragraph at the top, explaining what you are about to do. 
- The order in which you add the particles matters as WHFast is using Jacobi coordinates, Thus, adding Voyager before the Sun is problematic.
- If you want to resolve orbits of bodies around planets, WHFast is not a good choice. It will work if the timestep is very small, but you are better off with the general purpose IAS15 integrator or some other a specialized integrator that can handle moons.
- All the examples are automatically executed whenever someone makes a commit or creates a pull request. This is to check if any changes in the code break one of the examples. For that reason they should all run in a few seconds each so that all of them together take less than 30 minutes (which is the upper time limit on Travis CI). In your case querying the NASA database takes most of the time...

--
Hanno

PR #106, opened 2015-12-17
fixed badge links

dtamayo:
I just added some badges for REBOUNDx, and noticed that the links in the REBOUND badges weren't working because of an indentation problem.  I've fixed those, and added two more you're free to delete.  One auto-updates to give the current version on PyPI that I thought might be helpful to make sure we keep things in sync, and the other is a docs badge to rebound.readthedocs.org


hannorein:
I noticed that a while ago and couldn't figure out what was going on. Thanks a lot!

PR #107, opened 2015-12-17
Chris fixed bug in CloseEncounters.ipynb removing wrong particles

dtamayo:


hannorein:
Chris should get on github ;-)

dtamayo:
I should have given him that as an exercise you’re right.  We went through it together to learn a bit about how git works

PR #108, opened 2015-12-18
added dummy function for raising external exceptions

dtamayo:
I wanted to have a way for reboundx to raise exceptions when something goes wrong, so that you're not left wondering what happened with a Kernel died message in ipython.  After reb_integrate checks error codes, I have it call a dummy function that an external library can monkey patch.  Let me know if you think there's a better way to do this.  The corresponding reboundx branch that uses this is called 'effects' if you want to take a look

### DIFF
--- a/rebound/simulation.py
+++ b/rebound/simulation.py
@@ -806,7 +806,17 @@ class Simulation(Structure):
     @integrator_whfast_recalculate_jacobi_this_timestep.setter
     def integrator_whfast_recalculate_jacobi_this_timestep(self, value):
         self.ri_whfast.recalculate_jacobi_this_timestep = c_uint(value)
-    
+
+    def raise_external_exceptions(self, ret_value):
+        """
+        Dummy function that can be monkey patched by external libraries to check a simulation's
+        status field to raise custom exceptions, e.g.
+
+        if ret_value == 999:
+            raise CustomError("This or that kind of error occurred.")
+        """
+        pass
+
 # Integration
     def step(self):
         """
@@ -854,6 +864,7 @@ class Simulation(Structure):
                 raise Encounter("Two particles had a close encounter (d<exit_min_distance).")
             if ret_value == 4:
                 raise Escape("A particle escaped (r>exit_max_distance).")
+            self.raise_external_exceptions(ret_value)
         else:
             debug.integrate_other_package(tmax,exact_finish_time)
 


dtamayo:
Sorry, just realized the travis test failed, looking at it now

hannorein:
I'm not sure I understand. What is wrong with using throwing an exception? (I'm in today, if you are too)

hannorein:
I guess it's because the error occurs in c, not in python...

dtamayo:
I’m working from home today to catch our flight this evening.  The issue is that the errors come up in the C code, so just like the rebound C code sets r->status = REB_COLLISION and reb_integrate returns the code so that python can check and throw the corresponding exception, I wanted to add REBX_STATUS codes so that when they come up, they set r->status = REBX_ERROR_CODE so that they can be checked for when reb_integrate returns.

dtamayo:
so I have a REBX_STATUS enum that starts from 1001 to not collide with REB_STATUS, but does everything just like C.  This seemed like the simplest approach, but maybe there's a better one

dtamayo:
er, I meant it does everything just like REBOUND treats the error codes

hannorein:
Hm. Doesn't look very nice, but I do not have a better solution.

hannorein:
What kind of errors do you think you'll have to capture?

dtamayo:
I still get errors in modify_orbits_direct when there are close encounters, where the orbital element conversions give nans and the kernel just dies

hannorein:
One other solution: You could through a generic error in python and print something to stderr..

hannorein:
Or we could add a string to reb_simulation with the error message.

dtamayo:
I'm not quite sure how I would do that.  When I'm in a for time in times block calling sim.integrate, how would reboundx have control to raise an exception?

hannorein:
By setting r->status to an exception value.

dtamayo:
would i have to call sim.integrate and afterwards have the user always call something like rebx.check_errors?

dtamayo:
Oh, to a generic one in REB_STATUS, and printing the rebx one in stderr?

hannorein:
Yes, if you set `r->status = REB_EXIT_ERROR;` rebound should exit after this timestep. This causes an exception to be raised in python. If you want to tell the user why the exception happened, you'd need to either print it to stderr, or store a string in `reb_simulation` which python can then put into its exception.

dtamayo:
That would work...thanks!  I'll close this and open another pull request adding a custom_error_string in reb_simulation and have python check for whether the string is set when printing out the exception error.

hannorein:
Sounds good. It should also free the string when the simulation is freed or when a new error string overwrites the old one, to avoid a memory leak.

dtamayo:
I'm on plane about to leave now, but I think this may be complicated. I wrote a function to set the custom string, to make sure any previous one gets freed and new one gets allocated. However, when I pip install everything and run a python test where I call the string setter from reboundx (in the C code) I get a dlopen error saying it can't find the function. I think this is because when setup.py creates librevoundx.so, it doesn't know about librebound.so. Does that sound like it would be an issue?

hannorein:
I don't understand what the problem could be. I'd do something like (untested):

```c
void reb_break_with_custom_error(const chat* message){
   r->status = REB_EXIT_ERROR;
   if (r->custom_message){
      free(r->custom_message);
   }
   r->custom_message = malloc(strlen(message)+1);
   strcpy(r->custom_message, message);
}
```

dtamayo:
I implemented something very similar.  I think the problem is that setup.py for libreboundx knows nothing about where librebound.so is, so it can’t call REBOUND functions.  I guess the simple solution, like before, is for this function to be in REBOUNDx, though it seems strange for REBOUNDx to free REBOUND memory.  The solution I’d like to have is, through MANIFEST.in or something else there’s a way of including librebound.so in the pip installation of REBOUND, and have REBOUNDx find it.  Then I don’t have to duplicate code or mess with REBOUND’s memory from within REBOUNDx.  I’ll let you know where I get.

hannorein:
That doesn't make sense. There is nothing different about this one rebound function compared to all the others. Maybe make sure the tests are running with the right rebound version?

dtamayo:
I don’t think I’m calling any other rebound functions from within REBOUNDx.  I just use rebound.h for the structure definitions etc.  I did make sure I updated rebound.h with the set_custom_error_msg prototype and copied the new one into REBOUNDx.

hannorein:
Oh, wow. I didn't realize that. Then I need to think about it more.

dtamayo:
No worries.  It would be much better if I could have libreboundx find the rebound installation and have it include librebound.so so that REBOUNDx can call REBOUND functions.  I tried forever to get this working before, but I ran into a couple snags.  I posted a question on stack overflow that got no answers and gave up, but a couple weeks later I got an answer that might help.  I’ll give it a try again.

hannorein:
I don't think this PR is necessary anymore. If something goes wrong in REBOUNDx, then one can call `reb_warning()` which will raise an exception in REBOUND.

dtamayo:
Yup

dtamayo:
Thanks!

PR #112, opened 2016-01-15
Ari merge

silburt:
Here are all the changes, it's finally working well on my end minus the energy.

### DIFF
--- a/src/integrator_hybarid.c
+++ b/src/integrator_hybarid.c
@@ -116,6 +116,7 @@ void reb_integrator_hybarid_part1(struct reb_simulation* r){
         r->ri_hybarid.mini->additional_forces = reb_integrator_hybarid_additional_forces_mini;
         r->ri_hybarid.mini->ri_hybarid.global = r;
         r->ri_hybarid.mini->passive_influence = r->passive_influence;
+        r->ri_hybarid.mini->ri_hybarid.mini_active = 1; //for collision check in IAS15
     }
 
     // Remove all particles from mini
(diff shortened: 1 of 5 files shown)


hannorein:
Very good. One thing we need to talk about is the collision stuff. The current way is a bit of a hack. It should work by simply using the standard collision search routines plus a callback function.

src/integrator_hybarid.c
```diff
@@ -38,43 +38,7 @@
...
-        struct reb_particle* particles_mini = mini->particles;
-        struct reb_particle* particles_global = global->particles;
-        struct reb_particle* particles_global_prev = global->ri_hybarid.particles_prev;
-        
-        //forces from global into mini
-        double t_prev = global->t - global->dt;
-        const double timefac = (mini->t - t_prev)/global->dt;
-        for(int i=N_active;i<global->N;i++){    //planetesimals
-            if(global->ri_hybarid.is_in_mini[i]==0){             //find planetesimals which is part of global but not mini
-                const double ix = timefac*particles_global[i].x - (1.-timefac)*particles_global_prev[i].x; //interpolated values
```
hannorein: I think the problem was the minus sign here. Should have been a plus.

silburt:
OK. Let's talk about it next week. I'm going to try and see if I can get 
someone to cover me for the faculty meeting at St.G on wednesday so that 
I can come out to UTSC for our group meeting. Otherwise Thursday.

If we can figure out the energy before then, I can start running even 
more testing along with you and Dan (I've been testing all the way along 
with my code, but would be good for you and Dan to try and rip it apart 
at some point).

Ari

PR #122, opened 2016-01-27
added upgrading section to doc

dtamayo:


doc/python_quickstart.rst
```diff
@@ -119,3 +119,45 @@ Now, the time has advanced to t=1000. You can print out the particle positions w
...
+then simply::
+
+    pip install rebound --upgrade
+
+If this does not work, you have an old version of pip.  You can either upgrade pip (probably best!), or simply::
+
+    pip uninstall rebound
+    pip install rebound
+
+If you cloned the git repository, i.e., have a `rebound` directory on your file system, and are not too familiar with git, you can navigate to the `rebound` folder in the terminal and::
```
hannorein: I'm not sure if there is something such as too much documentation. But for a quick start guide, I would cut it here and not talk about how to use git.

PR #124, opened 2016-01-29
Add exit cond

dtamayo:
So a couple times now people have asked me to have a custom field in reb_simulation so that they can check for their own thing in heartbeat, like we do with exit_min_distance etc.  Many of these are probably not general enough to include in the master branch.  For Chris project, we'd like to keep track of distance from the star, and save a checkpoint if ever particles get close enough that tides might matter.  He's already using exit_min_distance for close encounters between planets, so we need an extra parameter like exit_min_peri.

Instead of just coding it in, I've created a script add_exit_cond.py that can do this generically.  The usage for this case is

python add_exit_cond.py "exit_min_peri" "double" "c_double" "101" "A particle got too close to the central body."

This will:
add a field in the reb_simulation struct with name "exit_min_peri"
add a field in the REB_STATUS enum REB_EXIT_MIN_PERI = 101 
add a placeholder in rebound.c's run_heartbeat if(r->exit_min_peri){ for you to fill in how you want to determine the exit condition.

add the corresponding _field_ in simulation.py
add an exception class "Exit_min_peri"
add a check for ret_value == 101 when calling reb_integrate from python that raises the new exception

I'm not sure whether you would find this useful, but thought I'd share in case.  If you don't, would you mind merging the typo I fixed in rebound.h and the line break in simulation.py (which my script needs to put in the new field?).

### DIFF
--- a/add_exit_cond.py
+++ b/add_exit_cond.py
@@ -23,7 +23,7 @@ with open("src/rebound.h", "w") as f:
         if "struct reb_simulation {" in rebh[i]:
             f.write("    {0} {1};\n".format(param_type_c, param_name))
         if "enum REB_STATUS {" in rebh[i]:
-            f.write("REB_{0} = {1},\n".format(param_name.upper(), param_enum_code))
+            f.write("    REB_{0} = {1},\n".format(param_name.upper(), param_enum_code))
 
 with open("src/rebound.c") as f:
     rebc = f.readlines()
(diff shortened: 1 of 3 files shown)


hannorein:
Probably best to talk about this in person. But why can't you just implement this in heartbeat without changing anything else?

    heartbeat(reb_simulation r){
        if (close_encounter){
            r->status = 101; 
        }
    }

dtamayo:
It’s probably better to talk in person, but in that case, wouldn’t the user would have to put logic into simulation.py ’s integrate function, like if ret_value == 101: do the specific thing you want to do?  By making a custom exception, they can still write their own code, and catch this new exception in a try except block like with encounters etc.

hannorein:
If sim.status is >0 it will simply exit as soon as possible. Then one can check afterwards what the status of sim was. 

If you want to use this from python, then I think it will not throw an exception, but one can just check sim.status.

hannorein:
It might be nice to have a generic exception for that purpose (that would just mean two extra lines in simulation.py).

dtamayo:
Yeah.  I think having python raise an exception is nicer than checking sim.status, since you can then catch it and do some custom thing.  I agree having a generic one would be useful.  Of course then someone will come along who wants to check for two custom things!  Let’s talk about it in person next week.

hannorein:
I don't see why two custom things would change this. After an exception is thrown the user can do whatever they want to.

dtamayo:
Wouldn’t they need two separate exceptions if they wanted to execute two different pieces of custom code depending on which exception gets thrown?

hannorein:
I'd do:

    try:
       for t in times:
          sim.integrate(t)
          do something
    except ParticleEscaped:
       print "particle escaped"
    except:
       print "something else happened"
       if sim.status==101:
          print "particle got to close to sun"
       if sim.status==102:
          print "aliens!"

dtamayo:
Ah, that makes sense.

hannorein:
I'll close this one. Let me know if you don't agree!

PR #126, opened 2016-02-09
Pand t

dtamayo:
I fixed a couple small bugs in the reb_M_to_E function that were causing weird behavior (tools.c).  I also added the ability to initialize orbits using the orbital period, which I've tested works as you'd expect when initializing stellar binaries/circumbinary planets.

I got tired of the nan stuff.  When I change every NAN to the C99 nan(""), I can no longer define const structs that are filled with nans because they aren't set at compile time.  I found a thread with people trying to deal with this (without a good resolution) spanning ~100 exchanges and several weeks, so I just put in ugly functions that returns structs with all nans.

Also, did you do anything so the diffs between ipynb's look nicer?  If so, it's really cool!


hannorein:
Thanks. I think the syntax is `nan(NULL)`. Is there any reason for `nan("")`? Otherwise I'll change it.

hannorein:
Apparently, 

    The call nan("") is equivalent to the call strtod("NAN()", (char**)NULL);.
    The call nan(NULL) is equivalent to the call strtod("NAN", (char**)NULL);.

I think the second one is nicer looking. For that reason only, I'll change it.

PR #127, opened 2016-02-09
fixed initial guess for hyperbolic orbits in kepler solver

dtamayo:
When writing an e-mail to Toni, I got a nan orbit back, and found the problem was that the Kepler solver would not always converge with the initial condition we had in there.  I updated it with the initial guess suggested by Danby, which converges for everything I've tried.


hannorein:
This also works for e=0?

PR #129, opened 2016-02-10
Hyperbolic orbits

dtamayo:
I added a HyperbolicOrbits.ipynb, and in the process updated plotting.py to do retrograde hyperbolic orbits right, as well as roughly get the scale for the tails right


hannorein:
I don't understand the difference between epoch_of_elements and epoch_of_observation. I would have thought epoch_of_observation is when the comet was observed and epoch_of_elements is when the elements are valid. But the way you describe it in the ipynb is the other way round. Are you sure this is correct?

PS: I moved the comets.txt data into the ipynb. I think it's not to have to load an external file for the example.

dtamayo:
makes sense on the file.  Maybe I got the epochs wrong…I’ll send Toni the ipynb and ask him to double check.

dtamayo:
You were right about the epochs.  I've updated the notebook and run it by Toni

hannorein:
Glad everything makes sense. (PS: I pushed a version with the orbits.txt dat in the ipynb file to hannorein/rebound/dtamayo-master. I'll merge it soon, just a little busy today)

hannorein:
Closing this one. See #131.

PR #133, opened 2016-02-22
Ari interpolate

silburt:
Calculating dE from collisions in global sim now.


hannorein:
Note that if you change the C structs, you might also want to change the python side of things.

PR #134, opened 2016-02-23
Ari interpolate

silburt:
Still need to update python packages for travis CI to not fail. I'll get that done today, hopefully this is the last commit where it will fail.


silburt:
Fixed a crucial bug that was leading to energy jumps in collision.c when the time step was very small.

PR #140, opened 2016-03-10
updated variation python doc

dtamayo:
I'm breaking up the changes I make into separate commit, thinking this might make it easier for you to review them.  Let me know if you would find it easier a separate way.  I'll keep adding to this pull request.

I deleted the below because I thought it was confusing that you say you get back a copy, but in the docstring example below you then modify it.  Of course you say this in the particles description below, but since the fields are undocumented and people shouldn't therefore change them, I thought the changes might be clearer.


hannorein:
Thanks. And yes, small pull requests make it easy to review.

PR #146, opened 2016-03-15
added heartbeat property

dtamayo:
same prototype as additional_forces, so I think quite straightforward.  Only perhaps tricky thing is that the function takes a Pointer to Simulation, so you have to do e.g.

def heartbeat(reb_sim):
    print reb_sim.contents.dt

### DIFF
--- a/rebound/simulation.py
+++ b/rebound/simulation.py
@@ -272,7 +272,20 @@ class Simulation(Structure):
     def post_timestep_modifications(self, func):
         self._ptmp = AFF(func)
         self._post_timestep_modifications = self._ptmp
-   
+ 
+    @property
+    def heartbeat(self):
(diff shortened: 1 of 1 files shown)


hannorein:
Note that there was an indentation issue.

hannorein:
Also, this works for me:

```
i=0
def h(sim):
    global i
    i=i+1
    print(i)
    
sim = rebound.Simulation()
sim.add(m=1)
sim.add(m=1e-3,a=1)
sim.heartbeat = h
sim.integrate(0.1)
```

dtamayo:
Sorry, I messed that up when trying to clean things up so you didn't get my old commits.  I just got something like that working.  Thanks for looking into it.

PR #156, opened 2016-04-06
added acc to operators

dtamayo:
need this to calculate acceleration of com using operators

### DIFF
--- a/src/tools.c
+++ b/src/tools.c
@@ -256,7 +256,7 @@ void reb_move_to_com(struct reb_simulation* const r){
 }
 
 struct reb_particle reb_get_com(struct reb_simulation* r){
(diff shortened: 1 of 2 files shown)


dtamayo:
probably for com I should be using calculate_com, so I updated that function to update the accelerations too

src/tools.c
```diff
@@ -256,7 +256,7 @@ void reb_move_to_com(struct reb_simulation* const r){
 }
 
 struct reb_particle reb_get_com(struct reb_simulation* r){
-	struct reb_particle com = {.m=0, .x=0, .y=0, .z=0, .vx=0, .vy=0, .vz=0};
+	struct reb_particle com = {.m=0, .x=0, .y=0, .z=0, .vx=0, .vy=0, .vz=0, .ax=0, .ay=0, .az=0};
```
hannorein: One could probably just to `={0.}`

PR #157, opened 2016-04-06
changed to {0.}

dtamayo:
I wasn't sure if this would create a warning because not every field in particle is a double (I had something like that when setting things to nan), but it doesn't seem to.

### DIFF
--- a/src/tools.c
+++ b/src/tools.c
@@ -256,7 +256,7 @@ void reb_move_to_com(struct reb_simulation* const r){
 }
 
 struct reb_particle reb_get_com(struct reb_simulation* r){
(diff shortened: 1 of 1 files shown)


dtamayo:
Also got a travis error that seems to say they don't have python 3.5 and couldn't download it?  Maybe related to the internet connectivity issues with Horizons?

"The command "curl -s -o python-3.5.tar.bz2 https://s3.amazonaws.com/travis-python-archives/binaries/$(lsb_release -is | tr "A-Z" "a-z")/$(lsb_release -rs)/$(uname -m)/python-3.5.tar.bz2" failed and exited with 18 during ."

hannorein:
I think the C99 standard says that it should do the right thing.

dtamayo:
Oh nice.  That’s definitely convenient…

hannorein:
Hm. Works for me, but yes, could be the connectivity thing.

PR #165, opened 2016-04-21
Ari clean

silburt:
some small fixes, +1 unit test, massive body collisions.

### DIFF
--- a/src/integrator_hybarid.c
+++ b/src/integrator_hybarid.c
@@ -125,6 +125,7 @@ void reb_integrator_hybarid_part2(struct reb_simulation* r){
         
         // Correct for energy jump in collision
         if(r->ri_hybarid.collision_this_global_dt && r->collisions_track_dE){
+            reb_move_to_com(r);
             double Ef = reb_tools_energy(r);
             r->collisions_dE += Ei - Ef;
         }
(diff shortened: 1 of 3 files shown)


src/integrator_hybarid.c
```diff
@@ -43,9 +43,6 @@ static void reb_integrator_hybarid_additional_forces_mini(struct reb_simulation*
 static void calc_forces_on_planets(const struct reb_simulation* r, double* a);
 double Ei;
 
-//temp!
-struct reb_particle ari_get_com(struct reb_simulation* r, int N_choice);
```
hannorein: probably want to remove that line.

hannorein:
I trust you the the remove function works. I can't easily figure out if it works in all cases. Maybe you can write couple of unit test where particles collide in the different cases for N_active.

silburt:
So far I've confirmed the following cases:
1) Single collision of massive-massive body, remains is just the two-body system.
2) Situation 1) but with an additional massive body orbiting further away.
3) Situation 2) but with an additional planetesimal also orbiting further away.
4) Situation 3) but the outer massive body + planetesimal are having a close encounter at the time of the massive-massive body collision. 

All seem to have the error scale appropriately. I'll see if I can come up with any additional tests as well as come up with a simple unit test.

PR #169, opened 2016-05-13
Second order derivative changes.

hannorein:
@dtamayo and @MrSwordFish, have a look at this pull request before I merge it into the master branch. It contains several changes regarding variational particles. Mainly:

- Includes derivatives with respect to the Pal coordinates that Rejean implemented. I removed the duplicates included in both classical and Pal coordinates. Only those involving `a` and `m` only are duplicate, all the others are unique.
- Derivates are now in their own file, `derivates.c`
- Unified syntax, all of them now take G, the original particle and the primary as an argument. Maybe the other way round would be more intuitive (G, primary, original particle)? **(Update: changed)**
- Better/simplified unit tests
- Cleaned up documentation (but would benefit from a bit more work).

A couple of comments:

- One thing I noticed (not sure if it's obvious) is that the pal derivatives keep `lambda` constant, whereas the classical ones keep `f` constant. This means we cannot (easily) use the pal `h` derivative instead of the chain rule if interested in h=e*sin(w). One just needs to stick to one coordinate system. A little confusing maybe, but I think that's just how things work out.
- We should at some point decide if we want to use `i` or `inc` as well as `l` or `lambda`. Seems a little inconsistent right now.


MrSwordFish:
I agree that (G, primary, original particle) is more intuitive. Or at least it is to me.

hannorein:
I merged it as a squash commit and will close this PR. Let me know if there are any other issues you come across.

dtamayo:
@hannorein, I’ve gone through the pull request (a big one!), and everything looks great to me.  I think the ipython examples are really clear.  And I agree about the different sets of coordinate systems—I don’t think there’s any good way around it.  People at low inclination will want to use the Pal coordinates, but people doing Kozai stuff will probably still want to use little omega (even if they could be using Pal coordinates).

Issue #160, opened 2016-04-13
Add a core utility reb_tools_am()?


meldonization:
to calculate total angular momentum of the system?

dtamayo:
This is added now as sim.calculate_angular_momentum() in python and reb_tools_angular_momentum() in C.

PR #171, opened 2016-05-19
Angmom

dtamayo:
Added function to calculate angular momentum.  3 in c for each component.  Python returns a list.  From https://github.com/hannorein/rebound/issues/160

### DIFF
--- a/rebound/simulation.py
+++ b/rebound/simulation.py
@@ -894,7 +894,16 @@ class Simulation(Structure):
         """
         clibrebound.reb_tools_energy.restype = c_double
         return clibrebound.reb_tools_energy(byref(self))
-    
+   
+    def calculate_angular_momentum(self):
+        """
+        Returns a list of the three (x,y,z) components of the total angular momentum of all particles in the simulation.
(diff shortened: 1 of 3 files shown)


dtamayo:
I tested that all three components are conserved in an IAS15 integration (phew!) ;)

hannorein:
I'm glad the angular momentum is conserved! ;-)

Super minor, but it looks a little repetitive. Would it make sense to have one function and a return value with type `res_vec3d` or one void function with arguments `struct res_simulation sim, double* Lxyz)`?

hannorein:
Merging it for now, feel free to change it later, if you think something like `reb_vec3d` makes sense...

hannorein:
Super-super minor comment: unit tests! ;-)

dtamayo:
Sure, I'm happy to do that.  In the code it said that reb_vec3d was only for internal use, so I didn't know if that was something you wanted to avoid exposing to the user.

hannorein:
I saw little use for it. Maybe this is one?

PR #172, opened 2016-05-20
Com

dtamayo:
added reb_get_com_range function that lets you specific first and last index to consider.  Works like python range function where 2,5 would give consider indices 2,3,4.  Rewrote jacobi_com and reb_get_com in terms of this function, and added an opposite to get_com_of_pair (get_com_without_particle), which is useful for me when stepping through jacobi coordinates backwards in reboundx.

also added functions for the particle operators, which python now calls.

Added separate group in C documentation, and added unit tests. 

One thing I wasn't sure about was whether divide_particle should check for whether the scalar value to divide by equals 0.


src/particle.c
```diff
@@ -160,3 +160,63 @@ int reb_remove_by_id(struct reb_simulation* const r, int id, int keepSorted){
...
+    p.vy = p1.vy - p2.vy;
+    p.vz = p1.vz - p2.vz;
+    p.ax = p1.ax - p2.ax;
+    p.ay = p1.ay - p2.ay;
+    p.az = p1.az - p2.az;
+    p.m = p1.m - p2.m;
+    return p;
+}
+
+struct reb_particle reb_add_particles(struct reb_particle p1, struct reb_particle p2){
```
hannorein: I've been very inconsistent myself, but I try to give functions a name such as `reb_FILENAME_...`. Do you think `reb_particle_add_particles` or `reb_particle_operator_add` or `reb_particle_add` would make sense?

src/tools.c
```diff
@@ -304,15 +320,27 @@ int reb_get_particle_index(struct reb_particle* p){
...
+        last = N_real;
+    }
+	for(int i=first; i<last; i++){
 		com = reb_get_com_of_pair(com, r->particles[i]);
 	}
 	return com;
 }
+
+struct reb_particle reb_get_com(struct reb_simulation* r){
+	return reb_get_com_range(r, 0, r->N);
```
hannorein: `N` or `N_real`? In python the default is `N_real`.

dtamayo:
Yeah that makes sense.  I think I prefer reb_particle_add rather than the clunkier reb_particle_add_particles.

hannorein:
The only thing is that is sounds somewhat similar to `reb_add()`

hannorein:
how about `reb_particle_plus`?

dtamayo:
I like that

dtamayo:
The check for N_real is in com_range

dtamayo:
I tried to make everything use com_range

hannorein:
Oh, I see. In what scenario would one trigger the check for N_real?

dtamayo:
As written reb_get_com passes N, so if there are variational particles com_range would set last to N_real rather than N.  It would also give you the COM for the whole system if you manually passed it a last that was too large (maybe that's not a behavior we want and get_com could pass N_real?)

hannorein:
I think the less complicated option would be to just pass `N_real` in `reb_get_com`, and have no check in `com_range`. When someone uses variational equations, using the com for all N_real particles is not really what one wants to use as the com for the variational particles.

dtamayo:
Makes sense

PR #173, opened 2016-05-20
Angmom

dtamayo:
Changed to returning vec3d.  Kept python as returning a list since it seems simpler, but maybe having an object where you can check L.x, L.y, L.z is more specific.  Any preference?

Added a unit test that makes sure L is conserved in a very short integration.


hannorein:
A list sounds better!

hannorein:
Thanks! This look much cleaner!

PR #174, opened 2016-05-21
fixed plotting.py

dtamayo:
The com change messed up the plotting routine.  I've fixed it and checked there are no other instances of calculate_com / get_com in the python and C code


dtamayo:
what do you think about setting the default to trails=True?  It looks much nicer and you can see the direction of the orbit.

hannorein:
Thanks!

If I remember correctly, there is a significant speed difference (trails are much slower). It could be that that's not the case anymore after the recent changes I made (easy to try out, just plot 100 random orbits with and without trails). Other than that, I agree, trails look nicer.

PR #175, opened 2016-05-21
changed default to trails=True

dtamayo:
Seems to take about the same amount of time, so changed default.  Output below

%%timeit -n 100

sim = rebound.Simulation()
sim.add(m=1.)
sim.add(a=1., e=random.random())
fig = rebound.OrbitPlot(sim)

100 loops, best of 3: 57.7 ms per loop

%%timeit -n 100

sim = rebound.Simulation()
sim.add(m=1.)
sim.add(a=1., e=random.random())
fig = rebound.OrbitPlot(sim, trails=True)

100 loops, best of 3: 56.9 ms per loop


hannorein:
Thanks. `OrbitPlot.ipynb` probably needs a small change too. I can do it, just not right now.

dtamayo:
I removed trails=True in the ipynb

PR #177, opened 2016-06-02
Hash

dtamayo:
This is not ready to merge.  I still have to update the documentation, examples, and clean up any references to id, but I think I added all the changes we talked about (except remove_by_hash), so I think other than that function this is the (proposed) final version of the code.  I'm not sure if it's useful to open a pull request early like this, but thought it might be useful for talking and tagging code.  

I've added a bunch of unit tests with which I tried to cover the scenarios I could think of.


rebound/__init__.py
```diff
@@ -57,11 +57,13 @@ class NoParticles(Exception):
     """The simulation exited because no particles are left in the simulation."""
     pass
     
-
+class ParticleNotFound(Exception):
+    """Particle was not found in the simulation."""
```
hannorein: Not super sure about the syntax. But doesn't this require a "pass" as for the ones above.

rebound/tools.py
```diff
@@ -0,0 +1,6 @@
+from ctypes import c_uint32, c_char_p
+from . import clibrebound
+
+def hash(string):
+    clibrebound.reb_hash.restype = c_uint32
+    return clibrebound.reb_hash(c_char_p(string.encode('utf-8')))
```
hannorein: Doesn't `reb_hash` take two arguments?

src/tools.c
```diff
@@ -925,3 +926,65 @@ void reb_tools_megno_update(struct reb_simulation* r, double dY){
...
+    hash *= 0x85ebca6b;
+    hash ^= (hash >> 13);
+    hash *= 0xc2b2ae35;
+    hash ^= (hash >> 16);
+
+    return hash;
+}
+
+uint32_t reb_hash(struct reb_simulation* const r, const char* str){
+    r->hash_ctr++;
```
hannorein: I *think* I would put the `r->hash_ctr++;` inside the if-statement. I would also check if `r==NULL`.

src/communication_mpi.c
```diff
@@ -73,7 +73,7 @@ void reb_communication_mpi_init(struct reb_simulation* const r, int argc, char**
     {
         struct reb_particle p;
         blen[bnum] 	= 1; 
-        indices[bnum] 	= (char*)&p.id - (char*)&p; 
+        indices[bnum] 	= (char*)&p.hash - (char*)&p;
```
dtamayo: I think this is OK, but wanted to run it by you.

dtamayo:
Could you check in particular the change in the line in communication_mpi.c?  I'm pretty sure it's right but wanted to run it by you.

dtamayo:
added separate pull request

PR #182, opened 2016-06-09
Suggestion about hash usage from python

hannorein:


dtamayo:
I agree that's a nicer syntax for get_particle

dtamayo:
I'll update the examples to reflect the syntax.

hannorein:
On thing to check is to make sure it works on both python 2 and 3.

dtamayo:
Does travis do python 2?

hannorein:
It checks both.

PR #184, opened 2016-06-09
updated uniquely identifying particles

dtamayo:


dtamayo:
Not sure this is the right way to contribute to a pull request, but anyway.  I always do everything in Python 3, and it seems to work, and travis passes, so I think this can be merged

PR #189, opened 2016-06-17
added sphinx flag to document __init__

dtamayo:
Note this will add documentation for all documented __init__ methods (which I think is what we want).  Currently Particle is the only class with documentation.  Doc now looks like this

screen shot 2016-06-17 at 8 52 26 am

### DIFF
--- a/doc/python_api.rst
+++ b/doc/python_api.rst
@@ -3,4 +3,5 @@ API documentation (Python)
 
 .. automodule:: rebound
    :members:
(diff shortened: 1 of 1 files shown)


hannorein:
Does that mean the documentation appears twice?

dtamayo:
No, the top is the docstring from just below class Particle (which appeared before).  This adds the documentation in __init__, which does not currently show up, so I think they're separate.

PR #191, opened 2016-06-17
modified Simulation.py for hermes so that mini.contents can be used.

silburt:


hannorein:
Are we sure this is not creating some circular reference now? I.e. `simulation` needs to know `ri_hermes` and vice versa?

hannorein:
Oh, I think it works. Because you put it before `simulation._fields_`...

hannorein:
In that case, I guess we could also change the `global` pointer?

silburt:
Yes, we had talked about making sure it was before simulation._fields_. We could also change global to a pointer.

PR #197, opened 2016-06-22
Adding access to particle coordinates as array

rjanish:
I've been using rebound through the python interface, and find myself often looping over a list of particle objects to construct a numpy array of particle positions, etc.  This gets pretty slow in python even for smallish particle numbers, and is glacial if you try to do this in a python heartbeat.  Instead of writing the full heartbeat in c, I've found it sufficient to build arrays of particle coordinates in c and then use numpy vectorized operations on those arrays in python.  

I've added a get_particle_data method to the python Simulation object which returns all of the particle data as arrays (coordinates, masses, velocities, radii and hashes), with the loop over particles done in c.  I thought this may be appreciated by users more familiar with python/numpy than c.


hannorein:
Thanks for the pull request! This looks very useful. Can I suggest a few changes:

- How about only returning one array, instead of four. Something in the form of: ((m,r,x,v,z,vx,vy,vz), (m,....)). 
- Remove the id/hash array. Unless you think there is a very good reason to keep it.
- I think the proper name for the function could be `reb_serialize_particle_data()`. 
- You also want to document the function in `rebound.h`
- You accidentally deleted two of the example problems.

rjanish:
Great!  Your suggestions sound good.  I am not sure how those examples disappeared, but I have re-added them.  I am fine renaming things, and thanks for catching the missing prototype in rebound.h (my c is a bit rusty).

As for the returned hashes, I have found them useful for removing particles.  If I use this function to get particle coordinates and want to remove particles based on the results of calculations with those coordinates, then it is nice to have the returned hashes to easily identify the particle corresponding to each coordinate.  The alternative would be identifying by array index, which seems like it would get messy after particles have already been removed - though perhaps there is another way to do it that I am not seeing.  Why do you not want to return the hash array?

I am also fine with condensing the output into one array.  I originally split them because I wanted the hash array to remain explicitly integer, and then figured if I was returning two arrays I might as well return five arrays.  But if we do keep the hashes we could always return just one int hash array and one float data array.

hannorein:
The only reason I thought it might be nicer without the hashes is that the hash is an integer whereas all the other variables are doubles. If they are all doubles, the array could easily be converted into one single numpy array. But I think you might have a good point with having the hashes. So I'm not sure what's better. 

Here's a more complicated proposal. Don't necessarily implement it right away, it's just an idea. What about a syntax like this:

    sim.serialize_particle_data("m", "x", "y", "z")

That could return four arrays m[], x[], y[], z[]. Or maybe

    sim.serialize_particle_data("m", "xyzz")

which returns 2 arrays. This could also include the id. Depending whether `sim.serialize_particle_data` receives a certain argument or not, one could save the memory allocation and potentially speed it up a bit, especially for very large particle numbers. E.g. if the argument "m" is not passed to `sim.serialize_particle_data`, then one could not allocated the space for the m array and simply pass NULL to the C function, which could then check if the pointer is NULL before accessing it, e.g.:

    if (mass){
        mass[i] = p[i].m;
    }

hannorein:
Added using a slightly different syntax in 61e94e6f7b7a52a0d65e18bb202a710ca6d74917 and 7984a1d55a8ce82c66566fb562d42fcc04a22df9

PR #198, opened 2016-06-22
reset lookup table parameters and pointers during binary read

dtamayo:
We don't write the lookup table to the binary file, so I think the right thing is to 0 / NULL out everything, and then the lookup table just gets remade the first time someone tries to get_by_hash

### DIFF
--- a/src/input.c
+++ b/src/input.c
@@ -136,10 +136,6 @@ struct reb_simulation* reb_create_simulation_from_binary(char* filename){
         r->allocatedN = r->N;
         r->tree_root = NULL;
 
-        // Reset lookup table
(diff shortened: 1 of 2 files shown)


src/input.c
```diff
@@ -136,6 +136,10 @@ struct reb_simulation* reb_create_simulation_from_binary(char* filename){
         r->allocatedN = r->N;
         r->tree_root = NULL;
 
+        // Reset lookup table
+        r->N_lookup = 0;
+        r->allocatedN_lookup = 0;
```
hannorein: I think this should also go in `reb_reset_temporary_pointers`.

PR #200, opened 2016-06-23
Think I solved the Issue

silburt:
I think this small change fixes the issue. Not sure if this somehow got changed back to "index" from 0 or if it was always set to "index", but I think it needs to be 0: 

If planet 1 has a close encounter with particle 4000 and planet 2 has a collision with particle 4, particle 4000 has been added to the global_index_from_mini_index array before particle 4, and wouldn't be depreciated if k starts at "index" instead of "0", but it needs to be depreciated. The issues that I showed you yesterday are no longer showing up, hopefully this resolved the issue.


hannorein:
Hm. I need to think about the logic at an earlier time of the day. 

You said you also saw the issue with IAS15, but this code only gets executed for HERMES.

Issue #208, opened 2016-07-05
contributing to docs


SeanBE:
Just wondering how you build rebound's docs locally. Wanted to see a change I made locally to one of the python method docstrings and it didn't pick it up.

I thought running `make doc` in the root dir would build the docs for both C++ and python source files.  
Am I missing something?

hannorein:
Yes. For the C part, you need doxygen installed.

SeanBE:
I've got all the dependencies. Looking at files under /doc/_build/html, it seems Sphinx didn't pick up my change. Build succeeds with 46 warnings. Is that what you're getting?

hannorein:
Oh, one more thing: You might also need jupyter as it converts ipython notebooks...

PR #209, opened 2016-07-05
update docstring for adding particle

SeanBE:
Check #208 .


hannorein:
Thanks! No pull request is too small!

PR #210, opened 2016-07-08
Particles

dtamayo:
Sorry I only got to this now.  Several things, and a few questions.  I think it's worth playing around with this a bit and get things right.  No rush on a response:

1)  I don’t think the workaround we discussed for allowing a hash of 0 quite works.  I think we’ll always be stuck with the fact that with the idiom in C of

struct reb_particle p = {0};

we can never know whether the user set p.hash to 0 on purpose or not.  I think the best solution (which I implemented and wrote into the C example) is for the user to always have to use reb_add_particle when using hashes (like we talked about requiring for using reboundx).  reb_add_particle now sets the hash by default to C_UINT32MAX.  **Does that sound good?** A couple options I like less:

a) set hash to C_UINT32MAX in reb_add and require the hash to always be set by the user after adding a particle to the simulation
b) disallow hash of 0

The add_particle solution I implemented has the issue you brought up in your e-mail:

star = reb_add_particle()
for (i in 0...10000):
   planetesimal = reb_add_particle()
   planetesimal-> x = star->x + 1.

would not work.  I give enough info in the C example to know this would not work, but maybe simplifying the C hash syntax (point 2) would make sense.

2) I think it is definitely confusing how many ways there are to access particles.  I removed all the by_name functions I had in C, so that you have to explicitly do things with hash and either set with or pass reb_tools_hash(hash) to make it clear that there aren’t separate hashes and names, only hashes.  It's a bit wordy.  **Should we replace reb_get_particle_by_hash with reb_get and reb_tools_hash with reb_hash?**  See the C example uniquelyidentifyingparticleswithhashes.  

e.g. reb_get_particle_by_hash(reb_tools_hash("Sun"))

3) I implemented the __del__ function for the particles mutable mapping in python so you can now do

del sim.particles[“earth”]

I think this is nicer/more consistent, since now you delete the particles in the same way you access them.  All these work:

del sim.particles[0]
del sim.particles[c_uint32(42)]
del sim.particles[“earth”]

**Should we add a deprecation warning to sim.remove and favor the above syntax?**.  With sim.remove it’s ambiguous whether you should do sim.remove(hash=42) or sim.remove(hash=c_uint32(42)), though I guess we could make it take both.  I made sim.add take both an int and a c_uint32 for hash.

For the del syntax above we have to choose a default for keepSorted.  True seems to be an advantage for lots of particles, only one particle moves in the array, so for most particles, r->particles[i] still points to the one you thought and when we add the sorting for faster access by hash, it doesn't have to get updated as often.  This poses a problem for few body systems though where we output orbital elements in Jacobi coordinates.  **So maybe keepSorted=1 is safer?**

4) I added new unit tests in test_hash.py and test_particles.py.  I added support for negative indices and slices in particles.  

5) I added a __repr__ = __str__ line in rebound.Particle so that when you do 

sim.particles[0]

in a jupyter notebook, it prints __str__ rather than ‘rebound.Particle at 0x2482349’ or something like that.

6) reb_get_particle_by_hash returns a NULL ptr if it doesn't find the particle.  I think it's good not to have an automatic warning/error since there might be cases where you don't know ahead of time if the particle will still be in so you want to check for NULL, but if the user mistakenly accesses it they'll get a seg fault and no idea why.  I tried to highlight this in the C example.  It's not an issue in Python.  **Is there a better way to deal with this with the messaging system you added?**


dtamayo:
Added qsort to lookup table update, and made search_lookup_table now scale as log N

hannorein:
I don't quite agree with 1). I don't see a problem with the hash=0 issue. Can you give me an example where you think what we had discussed earlier doesn't work?

I agree with 2). Regarding names, I would use `reb_get_particle_by_hash` and `reb_hash`.

3) I don't have a strong feeling about a deprecation warning on `sim.remove`. 

4) Nice.

5) Also nice.

6) I agree, a NULL pointer in C should be enough to indicate that a particle has not been found. In python, you probably want to throw an exception.

dtamayo:
1) How would it work if you add 4 particles with p={0} (so their hash is 0) and you wanted to set the third particle's hash to 0 so get_particle_by_hash(r, 2) returns you the third particle?

2) OK

3) I think we should have one or the other so it's not confusing.  I think maybe if we agree that there are cases where you'd want to keep sorted and others where not we should just keep remove and have __del__ pass. 

Could you remind me what you were suggesting for the string hashes? Flip the 1st bit to 1 in reb_hash and then when searching would you run through the lookup_table backwards linearly until the 1st bit is not 1 and then do a binary search with the remaining elements?

hannorein:
1) In that case you just return any particle with a hash of 0 since the user set all of them (knowingly or now) to 0.

3) I still don't have a strong feeling about this. I'd just keep remove() the way it is.

4) I don't remember why we were thinking about making the 1st bit one for all string hashes. Sorry, it's been too long!

dtamayo:
Closing this and opening a squashed pull request

PR #211, opened 2016-07-11
Added particles MutableMapping functionality + reb_add_particle

dtamayo:
I changed the default hash to 0, added some new tests of the sorting, and updated the examples.


dtamayo:
I also changed rebound.hash to also change int to c_uint32 so people import rebound.hash to access particles by numeric hash rather than from ctypes import c_uint32.  See the uniquelyidentifying... ipython_example.

hannorein:
Slowly going through this one.

Without `reb_add_particle`. Are the changes in particle.c still necessary?

dtamayo:
Thanks!  No, the changes at the top of particle.c aren't needed without reb_add_particle, just the lines starting from reb_search_lookup_table.  It's a bit of a mess in that I have another branch where I added the free_ap callback we talked about, and then fixed a bug in that the particles array wasn't being regenerated each time the particles MutableMapping is accessed, which in some cases can cause problems.  I have taken my other branch with free_ap and the fix, removed reb_add_particle and the other changes you referred to, and opened a separate pull request.  If we decide to add reb_add_particle afterward I can readd it.  I know you're just getting back (!), but if you have any time this week it would be great to talk about it in person since I'll be going to Cornell Mon-Wed.

PR #213, opened 2016-07-19
Added centimeter to units.py so rebound supports CGS

guillochon:
It would be great if CGS was natively supported in Rebound.


hannorein:
Thanks! (looks like you added grams, cm were already there)

guillochon:
Whoops! Yes I meant grams. Thanks for merging!

PR #216, opened 2016-08-08
Reorder

dtamayo:
This will only make sense after reading the e-mail I sent you a couple hours ago.  Moving the object_type member of rebx_effect did not help, because the whole structure is smaller than the first 9 elements of reb_particle (positions, velocities, and accelerations).  So since all these 9 quantities are changing every timestep, there's a good chance that in 10^10 timesteps you will match a given hash in all 10 places.  So I've moved the simulation pointer to the beginning of reb_particle.  This pointer won't change lots of times, so the chance of collision is negligible.  It meant that I had to also update the code and examples to initialize with {0} rather than {0.} since the first element is now a pointer.  I don't see any downside to this, but maybe you do.  I guess the alternative in REBOUNDx is another rewriting (?), or duplication of all the getter/setters for different types...I don't really like the latter since you might want to start adding parameters to other structures too in the future, and it will be a pain to have separate functions for each type, and for each variable type...

### DIFF
--- a/examples/variational_equations/problem.c
+++ b/examples/variational_equations/problem.c
@@ -13,7 +13,7 @@
 struct reb_simulation* create_sim(){
 	struct reb_simulation* r = reb_create_simulation();
     // r->integrator = REB_INTEGRATOR_WHFAST;  Only first order variational equations supported in WHFast.
-    struct reb_particle star = {0.};
+    struct reb_particle star = {0};
     star.m = 1;
     reb_add(r, star);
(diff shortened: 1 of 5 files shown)


hannorein:
I'm glad you found the issue. But let's talk about this in person. I'm not so keen on reordering the particle structure for two reasons:

- This is the fundamental data structure of REBOUND, so it should be changed only if absolutely necessary. 
- There was a reason to having all the `double`s at the beginning. The way there are no alignment issues. 
- It seems to (although only marginally) break people's code `{0.}` vs `{0}`

There are a couple of other ways I can think of to fix this. For example check if `particle->sim` is correctly set.

PR #218, opened 2016-08-17
HERMES examples

silburt:
Hanno,
Here are two ipython notebook examples which run fast (each under 10 seconds), and a longer C example of a planet migrating outward in a planetesimal disk over 10^5 years and beyond.

Ari


silburt:
Made the small change to the C example where I calculate the semi-major axis using the rebound tools.

PR #226, opened 2016-09-28
Fast simulation restarter

dtamayo:


dtamayo:
Added unit test for issue we fixed, and also a warning if post_timestep_modifications is set (I messed this up earlier without thinking and presumably others might too), along with a unit test for that.

rebound/simulation.py
```diff
@@ -413,7 +413,8 @@ def initSimulationArchive(self, filename, interval=None, interval_walltime=None)
         interval_walltime : float
             Interval between outputs in wall time (seconds). Useful for adaptive timesteps. 
         """
-        self.ri_whfast.safe_mode=0
+        if cast(self._post_timestep_modifications, c_void_p).value is not None:
```
hannorein: Why does it not work with PTM? Sorry if we already talked about it and I forgot.

dtamayo: You're right.  We talked at the beginning how you can't change things halfway through the simulation, but if you consistently apply the same ptm from the beginning it works (I just tested it to make sure).  I thought it didn't before, but this was because of the safe_mode=1 issue.  

I also realized I have to add calling the setup function and reboundx in simulationArchive's __init__ function, otherwise if you getSimulation from the middle of the simulation, those functions never get called.  But this means that if you load an archive and then getSimulation(0.), you call them twice (once in __init__ and then in getSimulation).  I'm not sure what the expected behavior should be...and didn't want to mess up the logic somewhere else...

PR #234, opened 2016-11-24
improved auto HSF routine and energy scaling for collisions

silburt:


src/integrator_hermes.c
```diff
@@ -119,9 +119,9 @@ void reb_integrator_hermes_part1(struct reb_simulation* r){
     
     reb_integrator_hermes_apply_forces(r, r->ri_hermes.a_i);
     
-    if(r->ri_hermes.mini_active && r->track_energy_offset){
-        r->ri_hermes.energy_before_timestep = reb_tools_energy(r);
-    }
+    //if(r->ri_hermes.mini_active && r->track_energy_offset){
```
hannorein: Do you think this can be removed rather than commented out?

PR #236, opened 2016-12-12
keep particles sorted

silburt:


hannorein:
Hm. I don't like that you override a variable normally set by the user. This sort of thing makes the outcome for a user impossible to predict. I think a warning message might be better.

silburt:
Makes sense. Turns out I forgot that I added a reb_exit command in particle.c a while back that already covers this issue (but exits the sim). So, you can either merge this version (which now doesn't exit the sim but prints a warning), or keep the old version and reject this pull request.

hannorein:
For the future, look into squashing your commits. Right now you have 2 changed lines but 5 commits...

PR #243, opened 2017-01-17
revamped autoHSF routine

silburt:
Ultimately, your idea of only applying the auto_HSF routine on (r_p1 - r_p2) < r_hill particles is already a condition occurring in the check_for_encounters routine. So instead of basically doing the same calculation twice, I think it's more efficient to merge the auto_HSF routine into the check_for_encounter. The essence is: if there's a close encounter add to mini, if there's no close encounter find the minimum encounter time (required to calculate the optimal HSF value). The tests I've done seem to be working fine. Give it a try and see if your examples work for you...


hannorein:
Ok. This seems to work better for some of these tests I've tried!

The way the autocalc_HSF routine is implemented right now, does it matter that one particle is a planet and the other a planetesimal? Could you swap them and get the same answer? If so, why is one called planetesimal and the other planet?

You're calculating the orbital elements around the total centre of mass. Why? Is that better than either using the star?

In all honesty though, this autocalc_HSF routine is still impossible to read or verify. I think you should spend an hour or two trying to refactor this. A couple of suggestions:

- Put some parts into smaller functions. 
- Always declare variables in the innermost scope possible, i.e.

```
double n;
if (){
   n = ...;
}
```
should be 
```
if (){
   double n = ...;
}
```
unless you need the variable after the if statement. This prevents bugs, makes it more readable and allows for optimizations.

- Several lines appear multiple times. If it's the exact same code, there is probably a better way to write it.
- Unit tests! If the functions are so complex, it's probably a good idea to write unit tests for all cases.

hannorein:
Here's an example of how I would write the unit tests. Create overlapping orbits. Then run the HSF auto routine on it to check if the routine is getting it right. Needs to be extended for all cases.

```
# Create two random particles with overlapping orbits.
sim = rebound.Simulation()
sim.add(m=1)
sim.add(a=1.)
sim.add(a=1.,l=np.pi)
sim.particles[1].vx += 0.5*(np.random.random()-0.5)
sim.particles[1].vy += 0.5*(np.random.random()-0.5)
sim.particles[2].vx += 0.5*(np.random.random()-0.5)
sim.particles[2].vy += 0.5*(np.random.random()-0.5)
dv = np.sqrt((sim.particles[1].vx + sim.particles[2].vx)**2 + (sim.particles[1].vy + sim.particles[2].vy)**2)
sim.integrate(2.*np.pi*np.random.random())
sim.particles[1].m = 1e-6
sim.particles[2].m = 1e-6
rhill = np.power(sim.particles[1].m/(3.*sim.particles[0].m),1./3.)

print("1/4 encounter timescale:", 0.25*rhill/dv)
print("timestep               :", sim.dt)
print("should have HSF > than :", sim.dt/(0.25*rhill/dv))


sim.testparticle_type = 1
sim.ri_hermes.hill_switch_factor = 3.
sim.ri_hermes.adaptive_hill_switch_factor = 1
sim.integrator="hermes"
rebound.clibrebound.reb_integrator_hermes_check_for_encounter(ctypes.byref(sim))
print("HSF returned by hermes :", sim.ri_hermes._current_hill_switch_factor)
```

silburt:
Hanno - I revamped the autoHSF routine so that it's hopefully much cleaner and easy to read. I've also done some testing and appears to still be working as it should. If you approve it, I'll start working on unit tests. If you still think it needs further cleaning I'd like to focus on that first before unit testing...

hannorein:
Awesome! I'll have a look! But I don't see a reason why you can't start writing some unit tests. I would write them on a relatively high level (see above). If we change something and they break, that's good! Then we know there is a problem...

hannorein:
One small comment. Correct me if I'm wrong. Right now, if there is no overlap between the two orbits nothing is calculated. But that means that two orbits which are just about to touch might be missed. Especially if one takes into account various approximations (we're not using the correct centre of mass). So I think there should be a small fudge factor, increasing/decreasing the maximum/minimum r by some number. I'm not sure what number makes sense. Maybe ~1.05?

silburt:
A hermes test failed... looking into it. 

I think you may be right but I need to think about it a bit more. Since the default HSF = 3, if a_1 = 1 and a_2 = 1.02, they will probably be recognized comfortably by IAS15, unless the two bodies are super small (e.g. m_1 = m_2 = 1e-9). If either of them have any kind of eccentricity the orbits will cross and the auto_HSF will kick in..

hannorein:
You're right this will only be important for particles with small Hill radii.

silburt:
Hanno,
Added some unit tests but they were a bit different from yours. Instead I decided to setup test cases such that CASE 1 and CASE 2 (inner/outer particle totally overlaps outer/inner particle) yield the same hill switch factor, and same thing for CASE 3 and CASE 4 (inner/outer particle partially overlaps outer/inner particle). 

The calculation you proposed might be a bit too simplistic, since particles with overlapping orbits probably have some eccentricity, making the max(v_r), max(v_phi) not equal to the instantaneous velocity. Plus, there's a bunch of ceiling and log functions to round to the nearest 1.25 hanging around. Maybe there's an ideal example for this type of test....

hannorein:
I went over the tests. I don't think these tests are sufficient. 

We set out to write a function that makes sure no particle ever travels more than 1/4 of the switching sphere within one timestep. That's what the tests should be testing. Right now they just test if the routine does anything. And the orbits are really extreme, but the edge cases are the crucial ones to test. We can talk about in person if it's unclear what I'm looking for.

silburt:
I added a general HSF check to the unit tests and it seems to be working well. The algorithm is called test_autoHSF_overlaping_orbits_general. It's very similar to the one you proposed above, except I place the planets close to each other so that the dv calculation is a bit closer to reality vs. them being separated by df = np.pi.

hannorein:
This looks good. You could put it in a loop and do that test 100 times for different random numbers.

hannorein:
Hm. Now I'm not so sure I understand the idea behind this:

```
HSF_approx = 3*sim.dt/(sim.ri_hermes.hill_switch_factor*rhillsum/dv)
self.assertGreaterEqual(sim.ri_hermes._current_hill_switch_factor,HSF_approx)
```

silburt:
The HSF_approx is always going to be an approximation, and depending on the specific setup, I was noticing that the following cases would show up:
case A: HSF_approx >~ sim.ri_hermes._current_hill_switch_factor, 
case B: HSF_approx <~ sim.ri_hermes._current_hill_switch_factor, 
case C: HSF_approx << sim.ri_hermes._current_hill_switch_factor. 

Since in the real auto_HSF routine we use a factor of 4 in the calculation (i.e. whats the required HSF such that a particle cannot travel 1/4th of a hill sphere in dt), I figured to allow for a bit of wiggle room in case A and case B I would have HSF_approx be a factor of 3 (i.e. whats the required HSF such that a particle cannot travel 1/3th of a hill sphere in dt), making it a slightly less stringent requirement, but still achieving the ultimate goal - testing that sim.ri_hermes._current_hill_switch_factor is responding to the orbital configuration and giving appropriate values.

PR #250, opened 2017-02-15
Transformations

dtamayo:
Refactored all the transformation routines from whfast and whfasthelio into separate transformations.c file. I added a separate documentation page for them in the c_api.

- I added simple jacobi mass calculation functions (that I used in REBOUNDx), but did not modify whfast to use them. 
- I renamed them to be more consistent and unambiguous now that they're all in one file
- to_jacobi_acc previously did not calculate ps[0].a since it wasn't needed, but I edited it to calculate it. I think it should be a trivial performance loss, and someone could get surprised (like me) when using it outside whfast
- I added a combination to_jacobi_posvelacc that's not used in REBOUND but I use in REBOUNDx

The changes I made since the last travis build don't have any effect within REBOUND, so it should still pass, but feel free to wait.


hannorein:
Thanks! This is very helpful and nicely separates the important WHFast part from the longish coordinate transformations. 

So you think there should be no changes to WHFast from this? E.g. it'll give the same results bit-by bit?

I think the files also need to go into the Manifest file.

dtamayo:
Yes, I only changed the function names, and added calculation of the jacobi particle 0 accelerations which are not used by WHFast. I just checked this branch and against master and get the same results bit by bit.

PR #252, opened 2017-02-28
working version

dtamayo:
These are the changes I need to make in REBOUND for the new REBOUNDx operator scheme to work. It adds a pre_timestep_modification function pointer analogously to post_timestep_modification, and calls it at the beginning of reb_step().

It should have no effect on code that doesn't use pre_timestep_modifications.

### DIFF
--- a/src/rebound.c
+++ b/src/rebound.c
@@ -69,6 +69,13 @@ const char* reb_githash_str = STRINGIFY(GITHASH);             // This line gets 
 void reb_step(struct reb_simulation* const r){
     // A 'DKD'-like integrator will do the first 'D' part.
     PROFILING_START()
+    if (r->pre_timestep_modifications){
+        reb_integrator_synchronize(r);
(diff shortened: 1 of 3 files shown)


src/rebound.c
```diff
@@ -69,6 +69,13 @@ const char* reb_githash_str = STRINGIFY(GITHASH);             // This line gets
 void reb_step(struct reb_simulation* const r){
     // A 'DKD'-like integrator will do the first 'D' part.
     PROFILING_START()
+    if (r->pre_timestep_modifications){
+        reb_integrator_synchronize(r);
+        r->pre_timestep_modifications(r);
+        r->ri_whfast.recalculate_jacobi_this_timestep = 1;
```
hannorein: Is that a good idea to set these flags here? It's not symmetric. I.e. not what we do for the post_timestep_modifications, where the user has to set the flags manually if desired.

PR #257, opened 2017-03-22
two remaining calls to whfast -> whfasthelio

rieder:
I could be mistaken, but all other calls to whfast seem to have been replaced with calls to whfasthelio...


hannorein:
I think this is correct. @silburt is still working on the final touches for HERMES.

PR #259, opened 2017-03-24
whfasthelio jump step now O(N)

silburt:
Hanno, 
Here is the code for the O(N) jump step. Looking at the Hernandez paper, it looks like the divisor of m[i]*p_h[i].v should be just m0, not (m0 + m[i]). This is reflected in the pull request. Maybe you can double check yourself and confirm. I'm not sure how (m0 + m[i]) snuck in there...


hannorein:
This looks good. I'll do a tiny bit of reordering...

PR #262, opened 2017-03-30
allow orbital elements for particle[0] if primary is provided

kecnry:
As far as I can tell, the only reason asking for the Keplerian elements for particle[0] is currently forbidden is because there is no way to determine wrt what primary particle.  However, everything seems to work so long as the primary particle itself is passed.


hannorein:
I think this makes perfect sense. Thanks!

PR #268, opened 2017-04-13
Simulation.add drops kwargs when particle is list

esemwy:
I ran across this while trying to simulate future planetary orbits from JPL Horizons.

`sim.add(['Sun', 'Earth', 'Jupiter'], date='2176-10-01 12:00')`

works on the current date rather than the one specified


hannorein:
I think that makes perfect sense. Thanks a lot!

Issue #269, opened 2017-04-19
Unable to turn off orbit trails/lines with OrbitPlot 


kdbertel:
It is possible that this is my using the wrong tool to accomplish what I want.

When I plot a simulation with rebound.OrbitPlot, all of my particles must display their orbit in some way, either as a trail or as a line for the orbit. What I would like is a way to turn that off, so that I can produce a graphic purely of the particles and their positions (but with all of the other plotting niceties that I get with OrbitPlot, instead of building my own MatPlotLib things up from scratch). One example of usage of this, for me, is to produce a series of images so that I can run through `ffmpeg` to produce a video of a system over time, where the orbit lines don't get distracting (especially if the orbits are unstable).

If this capability already exists, I am unable to find a way to do it.

hannorein:
Hi Keith!

The idea of OrbitPlot was to plot the orbits. If you just want to plot the positions, I think you can very easily build if up from scratch. Just store the positions in an array and plot is as a scatter plot. Plotting the orbits is the more complicated part that we were trying to help with with this class. 

Having said that, have you played around with the WebGL widget yet? It might be closer to what you want. I've just implement a way to save screenshots to a file which should allow you to make videos with ffmpeg. It's current on the `screenshot` branch. I hope to merge it into the master branch in a day or so. If you want to play with it, have a look here: https://github.com/hannorein/rebound/blob/screenshot/ipython_examples/ScreenshotsAndMoviesWithWebGL.ipynb


I hope this helps.
  Hanno

kdbertel:
Ah, that makes sense, though it'd be nice to quickly grab all of the `OrbitPlot` default ways of doing things, especially as I have essentially no experience or familiarity with MatPlotLib. I'll have to look through the source and see what I can pull out to do what I want.

I'm not working in iPython, so I'm not sure how much the WebGL stuff will help me, but it's really cool that it's being added.

hannorein:
Ah ok. If the source code is to confusing, let me know. I can easily add another flag to the class that does what you want.

kdbertel:
If you're fine with another flag in there, I'll see if I can't be ambitious and put together a pull request :)

hannorein:
That would be awesome!

PR #270, opened 2017-04-19
Update OrbitPlot to allow for not plotting orbit lines/trails

kdbertel:
This is a PR per the discussion in #269.

I added a new argument to `OrbitPlot` called `phases` that when `False` does not plot the orbit lines/trails.

The argument name was based on the internal name (`phase`) for plotting the orbits like that, but I'm sure there can be something better for it.

I was aiming to reduce the work done if `phases` was `False`, but I might have been too aggressive over what is behind the `if`; there was a bit of trial-and-error on my part for figuring out what did what.

### DIFF
--- a/rebound/plotting.py
+++ b/rebound/plotting.py
@@ -3,7 +3,7 @@ import math
 from .particle import Particle
 from itertools import cycle
 
-def OrbitPlot(sim, figsize=None, lim=None, limz=None, Narc=100, unitlabel=None, color=False, periastron=False, trails=True, lw=1., slices=False, plotparticles=[]):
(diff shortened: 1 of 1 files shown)


kdbertel:
Let me know if I screwed anything up; my day job is in iOS, and I only have so much familiarity with Python.

hannorein:
Ha! My day job is in Python and my hobby is iOS ;-) It looks good. I might only change the name to something other than phases...

kdbertel:
Totally reasonable, I just couldn't think of anything better.

hannorein:
I change it to `show_orbit`. Not much better, but at least somewhat self-explaining.

PR #305, opened 2017-11-06
Refactored Plotting

dtamayo:
After a couple iterations, I think I've broken out the essential functionality. It involved changing some of the logic around, but I've gone through it a couple times, plotted with the various permutations of the flags, and double-checked that OrbitPlot.ipynb renders the same plots as before with the new code. Made two helper functions with various options and put all the logic into OrbitPlot to replicate old behavior and keywords. Might be helpful to see first how these helper functions can be used to make custom plots:

[(https://github.com/dtamayo/systemsounds/blob/voiceover/projects/trappistvoiceover.ipynb)]

- Particle class now has a sample_orbit function that gives you back a list of xyz positions along the orbit. You choose Npts, how long to sample (defaults to reasonable values), and whether to sample in true or mean anomaly (OrbitPlot chooses for you based on whether hyperbolic like before)
- made it return a list rather than numpy array to keep particle.py numpy free
- I thought it made sense as a method on the Particle class since it's just giving you back numbers like all the orbital element properties

- fading_line just takes a list (or np array) and gives back a line collection linearly varying alpha
- made it accept **kwargs so it can take anything you want to pass that you could pass line collection


dtamayo:
Oops, branched from my HEAD. will make new PR

PR #306, opened 2017-11-06
refactored plotting

dtamayo:
After a couple iterations, I think I've broken out the essential functionality. It involved changing some of the logic around, but I've gone through it a couple times, plotted with the various permutations of the flags, and double-checked that OrbitPlot.ipynb renders the same plots as before with the new code. Made two helper functions with various options and put all the logic into OrbitPlot to replicate old behavior and keywords. Might be helpful to see first how these helper functions can be used to make custom plots:

[https://github.com/dtamayo/systemsounds/blob/voiceover/projects/trappistvoiceover.ipynb](https://github.com/dtamayo/systemsounds/blob/voiceover/projects/trappistvoiceover.ipynb)

- Particle class now has a sample_orbit function that gives you back a list of xyz positions along the orbit. You choose Npts, how long to sample (defaults to reasonable values), and whether to sample in true or mean anomaly (OrbitPlot chooses for you based on whether hyperbolic like before)

- made it return a list rather than numpy array to keep particle.py numpy free

- I thought it made sense as a method on the Particle class since it's just giving you back numbers like all the orbital element properties

- fading_line just takes a list (or np array) and gives back a line collection linearly varying alpha

- made it accept **kwargs so it can take anything you want to pass that you could pass line collection

### DIFF
--- a/python_examples/simple_orbit/problem.py
+++ b/python_examples/simple_orbit/problem.py
@@ -12,7 +12,7 @@ sim.add( a=1.4, e=0.1 )       # Massless test particle
 for o in sim.calculate_orbits(): print(o)
 
 # Output orbits in Heliocentric coordinates
-for o in sim.calculate_orbits(heliocentric=True): print(o)
+for o in sim.calculate_orbits(primary=sim.particles[0]): print(o)
 
 # Output cartesian coordinates
 for p in sim.particles: 
--- a/rebound/particle.py
+++ b/rebound/particle.py
@@ -470,6 +470,49 @@ class Particle(Structure):
             return clibrebound.reb_particle_divide(self, c_double(other))
         return NotImplemented 
 
+    def sample_orbit(self, Npts=100, trailing=True, timespan=None, useTrueAnomaly=True):
+        """
+        Returns a nested list of xyz positions along the orbit of the particle. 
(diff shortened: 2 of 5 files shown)


hannorein:
Awesome! Glad this is getting a revamp. Two questions:

- Shouldn't there be a primary=... when you create the temporary particles in the sample_orbit function? Right now it assume that the primary is at the origin?
- In OrbitPlotOneSlice you now loop over particles instead of orbits. This will be slow if there are many particles as the orbits need to be recalculated many times.

dtamayo:
Good points! I thought the runtime would be dominated by matplotlib overhead, but when I minimized the number of orbit calculations, it did turn out to speed it up by 30%. I've also added a primary keyword to all the plotting routines, keeping Jacobi as the default like it was before (I was wrong that it was sim.particles[0] before). I've also updated sim.calculate_orbits to take a primary for consistency with all the other functions, and raise an exception if heliocentric or barycentric keywords are passed. I've also checked that they don't show up in any of the ipython_examples.

dtamayo:
Oh, and rather than writing the C function for just the positions in python and putting sample_orbit in the Orbit class, I ended up just leaving it in Particle. I don't think it's inconsistent that you can now do ps[1].calculate_orbit() and ps[1].sample_orbit()

dtamayo:
Tests help!

hannorein:
Thank you!

PR #307, opened 2017-11-07
Realloc changes

bluescarni:
I noticed that simulations with lots of collisions tend to slow down over time. It seems it might be due to the fact that certain arrays related to collisions, when full, are reallocated to a size which is larger than the old size by a constant amount. Thus, as the number of total collisions increases, the reallocation becomes increasingly more expensive.

This PR changes the use of ``realloc()`` in two files so that reallocations double the original size. This results in a noticeable performance improvement on simulations with many collisions.


hannorein:
Thanks, that seems to be the right think to do!

PR #308, opened 2017-11-07
C++ support improvements

bluescarni:
This PR improves C++ compatibility for rebound. Specifically:

* it wraps the main ``rebound.h`` header with ``extern "C"`` linkage if a C++ compiler is being used,
* it uses a ``REBOUND_RESTRICT`` definition to wrap the ``restrict`` keyword. ``restrict`` is not part of C++, but it is supported in a few compilers (at least GCC and clang support it) with the alternative denomination ``__restrict__``. The ``REBOUND_RESTRICT`` definition expands to ``restrict`` if a C compiler is being used, to ``__restrict__`` if g++ or clang++ are being used, and to nothing otherwise.

### DIFF
--- a/doc/doxygen/Doxyfile
+++ b/doc/doxygen/Doxyfile
@@ -1980,7 +1980,7 @@ INCLUDE_FILE_PATTERNS  =
 # recursively expanded use the := operator instead of the = operator.
 # This tag requires that the tag ENABLE_PREPROCESSING is set to YES.
 
-PREDEFINED             =
+PREDEFINED             = REBOUND_RESTRICT=restrict
 
 # If the MACRO_EXPANSION and EXPAND_ONLY_PREDEF tags are set to YES then this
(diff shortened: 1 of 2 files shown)


hannorein:
I'm surprised there are compilers that do not support the restrict keyword. It's part of the ISO C99 standard. Is there a specific one that you've come across and this is an issue?

bluescarni:
My use case at the moment is that I am compiling the rebound shared library using a C99 compiler (gcc), and linking it to a C++ codebase. In order for the linking to work, the rebound header ``rebound.h`` has to be parsed by a C++ compiler (g++ or clang++) which does not understand the ``restrict`` keyword, as ``restrict`` is not part of any C++ standard so far. E.g., see:

https://gcc.gnu.org/onlinedocs/gcc/Restricted-Pointers.html

hannorein:
Thanks for that. I'll merge this in.

PR #319, opened 2018-01-06
Raise

dtamayo:
added reb_collision_resolve_halt as default when collisions are all. Simply sets r->status = REB_EXIT_COLLISION silently and halts the integration. In Python it will raise a Collision exception. Added to the end of CloseEncounters ipynb how to use it and set up a custom python function.

This changes the default from hardsphere, so I've gone through all the C examples and the SaturnsRings ipynb and set collision_resolve to hardsphere explicitly in ones where collision_resolve was not already set. I also changed shearing_sheet_2, which defines a custom resolve function in problem.c but I think was using the default hardsphere because it wasn't setting collision_resolve.

In doing that I realized that the C examples still mostly have tabs, so I went through and changed all tabs to spaces. It's a pain in the side by side viewer...I separated it out into a separate commit so you might view the other commits individually to see the changes that actually matter.

I also added -Wno-unused-result to Makefile.defs (only on Linux), because on gcc (e.g. in the travis build) we get a lot of obnoxious warnings (for each of the C examples) on not using the return values from fread. I think checking the return value for each fread would make the code really ugly, and I would think it's fine not to add them (famous last words!) given that we control how the binaries are written and have unit tests.

### DIFF
--- a/examples/bouncing_balls/problem.c
+++ b/examples/bouncing_balls/problem.c
@@ -11,33 +11,33 @@
 #include "rebound.h"
 
 int main(int argc, char* argv[]){
-	struct reb_simulation* r = reb_create_simulation();
-	// Setup constants
-	r->integrator	= REB_INTEGRATOR_LEAPFROG;
-	r->gravity	= REB_GRAVITY_BASIC;
-	r->collision	= REB_COLLISION_DIRECT;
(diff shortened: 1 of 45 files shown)


hannorein:
Thanks Dan! This is awesome!

Issue #331, opened 2018-03-23
Probably bug with N_active and variational particles


Kubala95:
I would like to share with my doubts about calculation of gravitational acceleration. This line looks invalid for me:

`const int _N_active = ((N_active==-1)?N:N_active) - r->N_var;` (number 67, src/gravity.c)

If the second option is true, decreasing it by N_var isn't needed and is even harmful - N_active is precisely last index representing massive particle, if N_var > 0 the result will be completely another than user wish. In the case of use tangent map in WHFast, the N_active != -1 will totally disable gravity.
I think more correctly would be:

`const int _N_real = N - r->N_var;`
`const int _N_active = ((N_active==-1)?_N_real:N_active);`

Greetings

dtamayo:
Thanks for this catch. I agree this is a bug. The following code gives unphysical results

    sim = rebound.Simulation()
    sim.add(m=1.)
    sim.add(m=1e-3, a=1)
    sim.add(a=2)
    var_da = sim.add_variation() 
    sim.N_active = 2
    sim.integrate(100.)

but runs perfectly fine if you take out either the variations or setting sim.N_active. Do you want to send us a pull request so that you get credit for the fix?

Kubala95:
Thanks for answer. It will be very nice to me if I could send you this small pull request. :)

PR #332, opened 2018-03-26
Fixed bug with 'N_active' field and variational particles

Kubala95:


dtamayo:
Thanks! This looks great to me, and looks the failed unit test is due to the inability of one of the Travis threads to install one of the dependencies and not with the change. 

I'll let @hannorein take a look before merging it. He's been away but I think gets back today.

hannorein:
I think this makes sense. Well done for spotting this and thanks for the pull request!

hannorein:
Closes #331.

PR #336, opened 2018-04-24
added manual snapshot append functions

dtamayo:
I moved the time information into reb_simulationarchive_append and added a python function that lets you manually append snapshots. I think it gives the intended behaviour. Here's an example that gives extra snapshots at 10500 and 20700 as expected. 

```
import rebound
import numpy as np

sim = rebound.Simulation()
sim.add(m=1.)
sim.add(m=1e-3, a=1.)
sim.add(m=1e-8, a=2.)
sim.move_to_com()
sim.dt = sim.particles[1].P*0.05  # timestep is 5% of orbital period
sim.integrator = "whfast"

sim.initSimulationArchive("satest.bin", interval=1.e3)
sim.integrate(1.05e4)
sim.remove(2)
sim.simulationarchive_append()
sim.integrate(2.07e4)
sim.simulationarchive_append()
sim.add(m=1.e-8, a=5.)
sim.integrate(3.e4)

sa = rebound.SimulationArchive("satest.bin")
for sim in sa:
    print(sim.t, sim.dt)
```

### DIFF
--- a/rebound/simulation.py
+++ b/rebound/simulation.py
@@ -480,7 +480,9 @@ class Simulation(Structure):
         if interval_walltime:
             self.simulationarchive_interval_walltime = interval_walltime
 
-    
+    def simulationarchive_append(self):
+        clibrebound.reb_simulationarchive_append(byref(self))
+
     def process_messages(self):
         clibrebound.reb_get_next_message.restype = c_int
         buf = create_string_buffer(c_int.in_dll(clibrebound, "reb_max_messages_length").value)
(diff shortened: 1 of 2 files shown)


dtamayo:
To be more explicit...it means that it will not mess around with the interval you've preselected if you manually add a snapshot, so you get snapshots at 10,000 and 11,000 despite adding a snapshot at 10,500. But if you specify wall time, then if you add a snapshot manually, it will update the time so you will only get a new snapshot automatically afterwards after the interval_walltime elapses. That behaviour made sense to me, but maybe I'm not thinking of everything!

hannorein:
Thanks! Make sense! Sorry for the mess on this branch! But I think it's slowly converging...

PR #337, opened 2018-04-26
fixed index bug and updated python syntax

dtamayo:
I think this fixes a bug where the last snapshot was not being read in a simulationarchive. Adding snapshots manually from python works great!

### DIFF
--- a/rebound/simulation.py
+++ b/rebound/simulation.py
@@ -466,8 +466,8 @@ class Simulation(Structure):
             clibrebound.reb_simulationarchive_automate_walltime(byref(self), c_char_p(filename.encode("ascii")), c_double(walltime))
         self.process_messages()
 
-    def simulationarchive_append(self):
-        clibrebound.reb_simulationarchive_append(byref(self))
+    def simulationarchive_snapshot(self, filename):
+        clibrebound.reb_simulationarchive_snapshot(byref(self), c_char_p(filename.encode("ascii")))
 
     def process_messages(self):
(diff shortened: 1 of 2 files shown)


dtamayo:
Oops...don't merge yet. It now fails some tests.

dtamayo:
Hm...I'm pretty sure the fix is needed, but it seems like all the whfast integrators with safe_mode = 0 fail when this change is made. I can't figure out why. 

Without this fix, then the following code

```
import rebound
sim = rebound.Simulation()
sim.add(m=1.)
sim.add(m=1e-3, a=1.)
sim.automateSimulationArchive("satest.bin", interval=1.e3, deletefile=True)
sim.integrate(1.01e4)
sa = rebound.SimulationArchive("satest.bin")
print(sa[-1].t)
```

prints 9000.158524373, rather than ~10,000 as expected.

hannorein:
This was an issue with the unit tests, not the code.

If we run the simulation with 

sim.automateSimulationArchive("ss.bin",10.)
sim.ri_whfast.safe_mode = 0
sim.integrate(40,exact_finish_time=0)

then a shapshot gets crearted at 40.0012 where WHFast is synchronized. If we run the simulation from 0 to 80 in one integrate() call, then it will not get synchronized at 40. 

I think this was different in the old version where it would always save the unsynchronized coordinates when the safe mode is turned off. I think the new way makes more sense, but I need to think about it.

PR #343, opened 2018-05-08
fixed reb_exit and whfast com variations

dtamayo:


hannorein:
Thanks a lot! 

I think the logic makes sense!

PR #344, opened 2018-05-08
added close function to simulationarchive

dtamayo:
You can get an error for having too many files open in Jupyter, so I added a __del__ function following what we do for Simulation. I also moved the sim archive functions in rebound.h into the right doxygen section (they weren't getting picked up in the docs)

### DIFF
--- a/rebound/simulationarchive.py
+++ b/rebound/simulationarchive.py
@@ -86,6 +86,10 @@ class SimulationArchive(Structure):
         self.tmin = self.t[0]
         self.tmax = self.t[self.nblobs-1]
 
+    def __del__(self):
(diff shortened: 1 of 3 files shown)


hannorein:
Good point. Thanks!

(I think this should be an internal function though. Otherwise it might be confusing having both a "close" and "free" function for the simulation archive.)

dtamayo:
Makes sense. I'll update

dtamayo:
Hm, I think the only issue is that it needs to be accessible to python. I think I've set things up exactly the same as reb_free_simulation and reb_free_pointers

hannorein:
I agree. I'd just not add the documentation (or make it private).

PR #345, opened 2018-05-08
made free_pointers functions private in doc

dtamayo:
made free_pointers functions private in doc for both simulation and simulationarchive


dtamayo:
sorry didn't see the not worth updating!

Issue #355, opened 2018-07-30
OrbitPlot doesn't render hyperbolic orbits correctly


hannorein:
Not sure why...

```
%matplotlib inline
import rebound
sim = rebound.Simulation()
sim.add(m=1)
sim.add(m=1e-3,a=1)
sim.add(m=1e-3,a=2,e=0.4)
sim.add(m=1e-3,a=-2,e=1.4)
sim.integrate(0.4)
```

![unknown](https://user-images.githubusercontent.com/53376/43429599-40d004c6-9432-11e8-9984-4a4413472bd3.png)

Issue #340, opened 2018-05-05
Potential bug in reb_tools_M_to_E


pmplewa:
Hello! I think there could be a bug in the function `reb_tools_M_to_E(e, M)`. For high e and negative M, I sometimes get wrong values for E. For example, when I fix e at 0.96 and let M vary from -2pi to 2pi in small steps, it happens in several cases:

![example](https://user-images.githubusercontent.com/19286814/39663674-e2195be2-5077-11e8-9902-2a24c7a98553.png)

As one specific example, setting e = 0.96 and M = -2.0106192982974678 results in E = -4.959593. I was only able to get the expected result, E = -2.547771, after changing the function like so:

```C
double my_M_to_E(double e, double M){
  double E;
  int M_is_positive = M >= 0. ? 1 : 0;
  M = fabs(M);
  if(e < 1.){
    E = e < 0.8 ? M : M_PI;
    double F = E - e*sin(E) - M;
    for(int i=0; i<100; i++){
      E = E - F/(1.-e*cos(E));
      F = E - e*sin(E) - M;
      if(fabs(F) < 1.e-16){
        break;
      }
    }
  }
  else{
    E = M;
    double F = E - e*sinh(E) + M;
    for(int i=0; i<100; i++){
      E = E - F/(1.0 - e*cosh(E));
      F = E - e*sinh(E) + M;
      if(fabs(F) < 1.e-16){
        break;
      }
    }
  }
  E = M_is_positive ? E : -E;
  E = mod2pi(E);
  return E;  
}
```

Can you reproduce this problem? If so, what's the best way to fix it? Thanks for checking!

hannorein:
Thanks for the bug report! Yes, I can confirm the behaviour. I used the `reb_tools_M_to_f` function `(which calls the `reb_tools_M_to_E` function) and get the same issues for negative `M`. Let me stare at the function for a little to figure out what's going on...

![unknown](https://user-images.githubusercontent.com/53376/39676613-b1c38890-513b-11e8-8de4-1d5b494ca5c7.png)


Code to reproduce the plot: 

```
import rebound
import ctypes
%matplotlib inline
import matplotlib.pyplot as plt
import numpy as np
rebound.clibrebound.reb_tools_M_to_f.restype = ctypes.c_double
fs = np.linspace(-2.*np.pi,2.*np.pi,1000)
Ms = np.zeros(len(fs))
for i in range(len(fs)):
[...]

PR #356, opened 2018-07-31
negative M values

dtamayo:
Fix for issue #355 

I've narrowed down the issue is with initializing particles using negative M values. This pull request gives back the right orbit. Alternatively if you use the current master and try:

```
import rebound
sim = rebound.Simulation()
sim.add(m=1)
sim.add(m=1e-3,a=-2,e=1.4, M=-0.001)
sim.particles[1].x
```

you'll get a jump when you switch M from negative to positive. 

I'm just a bit leery that fix comes from commenting out a line that itself has a comment that it's there to avoid numerical artifacts for negative numbers!! Could you take a look @hannorein ?

### DIFF
--- a/src/tools.c
+++ b/src/tools.c
@@ -424,7 +424,7 @@ static double mod2pi(double f){
 }
 
 double reb_tools_M_to_E(double e, double M){
-    M = mod2pi(M); // avoid numerical artefacts for negative numbers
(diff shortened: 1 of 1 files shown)


hannorein:
Thanks for spotting this. It's related to #340. I need to think about a good way to fix this...

PR #378, opened 2019-03-30
Lyap

dtamayo:
Couldn't figure out why the lyapunov times in the stability project integrations weren't making sense. After chasing down the wrong things for a while I now understand the MEGNO much better :) It turned out it was a bug with calculating the megno with WHFast with keep_unsynchronized = 1. synchronize() caches the p_j and at the end resets the p_j back, but at the end of part2, the variational particle calculation uses the p_j, which have been reset to mid-timestep values. I think the new line fixes the problem, but I also wondered whether it would make sense to have a dedicated p_j_cache array in the simulation structure so that we don't malloc and free one every timestep. In that case the new line I added wouldn't be necessary.

I also added a couple things I ran into while debugging:

- I would find it useful to be able to get exactly the same megno/lyapunov times every time for the stability stuff, so I added a seed option to megno_init(). If you don't like it you can just remove the last commit.

- I made integrator_reset() callable from Python

- This is more general...some errors can make the code hang. For example, 

```
sim = rebound.Simulation()
sim.integrator = "whfast"
sim.ri_whfast.coordinates="whds"
sim.add(m=1.)
sim.add(m=1.e-3, a=1.5, e=0.1, inc=0.1)
sim.init_megno()
sim.integrate(2)
```

In whfast_init it will reb_error because you can only use variational particles with jacobi coordinates. Then in integrate's reb_check_exit(), this condition passes ```if (r->status>=0 || reb_error_message_waiting(r))```, the function returns without updating the status, so whfast_part_1 gets called again, which does nothing because it reb_errors, and just loops forever. 

It seems to me like the different error statuses were from the time we needed them before you set up the error logging system. Maybe this is a naive solution, but I thought that one could raise reb_errors anywhere in the code and then have reb_check_exit catch them in the above condition and change the status to 1, and rather than have python translate the reb_status, it could just process messages. I'm probably missing all sorts of edge cases!

### DIFF
--- a/src/integrator_whfast.c
+++ b/src/integrator_whfast.c
@@ -788,9 +788,11 @@ void reb_integrator_whfast_part2(struct reb_simulation* const r){
             struct reb_particle* const particles_var1 = particles + vc.index;
             const int index = vc.index;
             // Centre of mass
+            reb_transformations_inertial_to_jacobi_posvel(ri_whfast->p_jh+index, particles_var1, particles, N_real);
             ri_whfast->p_jh[index].x += r->dt/2.*ri_whfast->p_jh[index].vx;
             ri_whfast->p_jh[index].y += r->dt/2.*ri_whfast->p_jh[index].vy;
             ri_whfast->p_jh[index].z += r->dt/2.*ri_whfast->p_jh[index].vz;
+            reb_transformations_jacobi_to_inertial_posvel(particles_var1, ri_whfast->p_jh+index, particles, N_real);
             if (r->calculate_megno){
                 reb_calculate_acceleration_var(r);
                 const double dx = particles[0].x - particles[1].x;
--- a/src/integrator_whfast.c
+++ b/src/integrator_whfast.c
@@ -788,7 +788,9 @@ void reb_integrator_whfast_part2(struct reb_simulation* const r){
             struct reb_particle* const particles_var1 = particles + vc.index;
             const int index = vc.index;
             // Centre of mass
-            reb_transformations_inertial_to_jacobi_posvel(ri_whfast->p_jh+index, particles_var1, particles, N_real);
+            if (ri_whfast->keep_unsynchronized){ // unsync reverts p_j in synchronize(), so have to transform back
+                reb_transformations_inertial_to_jacobi_posvel(ri_whfast->p_jh+index, particles_var1, particles, N_real);
+            }
             ri_whfast->p_jh[index].x += r->dt/2.*ri_whfast->p_jh[index].vx;
             ri_whfast->p_jh[index].y += r->dt/2.*ri_whfast->p_jh[index].vy;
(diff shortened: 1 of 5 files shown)


src/rebound.c
```diff
@@ -572,6 +572,9 @@ int reb_check_exit(struct reb_simulation* const r, const double tmax, double* la
     }
     const double dtsign = copysign(1.,r->dt);   // Used to determine integration direction
     if (r->status>=0 || reb_error_message_waiting(r)){
+        if(r->status <= 0){
+            r->status = 1;
+        }
```
hannorein: I'm not sure I understand the consequences of this change.

src/integrator_whfast.c
```diff
@@ -789,6 +788,7 @@ void reb_integrator_whfast_part2(struct reb_simulation* const r){
             struct reb_particle* const particles_var1 = particles + vc.index;
             const int index = vc.index;
             // Centre of mass
+            reb_transformations_inertial_to_jacobi_posvel(ri_whfast->p_jh+index, particles_var1, particles, N_real);
```
hannorein: Is this doing extra work when `keep_unsynchronized=0`?

src/rebound.c
dtamayo: What was happening with the code snippet above is that r->status = -1, reb_error_message_waiting(r) would trigger because whfast reb_errored on not using Jacobi coordinates, so check_exit would just return with r->status = -1. Then whfast would return immediately on whfast_init on the reb_erorr without doing anything and it would loop. So I thought one solution (perhaps naive) is that if there is a reb_error_message_waiting, we could just set reb_status to the generic error = 1. Travis passes, but I might be missing some logic/edge cases...

src/integrator_whfast.c
dtamayo: good point!

src/rebound.c
dtamayo: I wasn't sure if that could be the default behavior so that we don't have to treat and translate positive r->status values explicitly in python when sim.integrate returns and could just process_messages to raise whatever reb_errors occurred in C

dtamayo: I see now that you didn't write that you didn't understand why it was there, but what the consequences would be...travis passes, but I agree it's a major change and quite unclear across all cases!

hannorein: Thanks. I think this is doing the right thing. Although I would probably write it like this:

```
if (reb_error_message_waiting(r)){
    r->status = REB_EXIT_ERROR;
}
if (r->status>=0){
        ...
```

for readability.

dtamayo: Good idea! Thanks

hannorein:
Thanks a lot for figuring this all out!

dtamayo:
Feels better when I don't have to just dump my problems on your plate :) Thanks for the quick improvements!

PR #379, opened 2019-04-04
Newlyap

dtamayo:
So I fooled myself into thinking the previous patch worked because the tests were all for a single planet. Looking at it for several planets, the patch didn't make sense. What we need when unsynchronized is to actually synchronize the p_j, so we have to do what synchronize() does, then update the megno, and only revert back at the end of that. This is a hacky way to do that, but it works. Probably the better one would be to have a separate unsynced_p_j array in the simulation structure that is always what gets used when restarting the integration, and the others are just for output. But I haven't got into the details of that, so I'm probably missing all sorts of complications in the logic that that might cause. I think this is a reasonable patch that only does extra work when keep_unsynchronized=1. 

I also updated test_megno.py to have multiple planets and have both closely packed regular cases and chaotic ones. Sorry for the mixup!

### DIFF
--- a/rebound/tests/test_megno.py
+++ b/rebound/tests/test_megno.py
@@ -13,7 +13,8 @@ class TestMegno(unittest.TestCase):
         self.sim.integrator = "ias15"
         self.sim.add(m=1)
         self.sim.add(m=1e-3,a=1.5,e=0.1,inc=0.1)
-        self.sim.init_megno()
+        self.sim.add(m=1.e-3, a=15., e=0.1, inc=0.1)
+        self.sim.init_megno(seed=0)
         self.sim.integrate(1000)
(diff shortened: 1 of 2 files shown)


hannorein:
Oh wow. That sounds like a mess. I'll guess I'll merge this in, but honestly, I don't really understand the consequences...

dtamayo:
lol. It should be very much self-contained! There are no consequences whatsoever unless keep_unsynchronized=1, and when it is, it's doing something very simple.

It needs to do exactly what synchronize() does (half kepler half com step), but needs to wait to revert the p_j back to the middle of the step until after it does the megno calculation. I could have copy-pasted what's in synchronize() into the megno part, but this seemed a little cleaner (manually cache and restore the p_j, and just trick synchronize() into not reverting by toggling the flags. Thanks!

PR #383, opened 2019-05-30
made keep_unsync incompatible with safe_mode

dtamayo:
Found another edge case...I don't think keep_unsynchronized should be compatible with safe_mode. There's a bug in the current version where safe_mode overwrites the p_jh when keep_unysynchronized=1 because it keeps seeing that is_synchronized==0. 

That could be fixed, but I think the better solution is to say you can't have both since they do opposite things. safe_mode wants to synchronize every timestep, keep_unsynchronized wants to avoid that. I thought that a better option than silently setting safe_mode to 0 when you set keep_unsynchronized=1 is to raise an error explicitly telling you you need to set safe_mode = 0 if you want to use the keep_unsynchronized flag, but maybe there's a better way. 

I also found out the hard way that the C rand() function is not machine independent (makes sense), which makes my earlier patch of being able to pass megno_init a seed less useful. Given how much effort we've put into reproducibility, would it make sense to try to fix this? I thought that one option might be to use the hash function we already have in there. Though I guess some of the randomization in REBOUND does want good RNG properties? Maybe not worth the effort? In any case, I had to increase the length of the megno test integrations, because for some RNG draws, the previous lengths didn't make it through the early transients to reach a values of 2.

### DIFF
--- a/rebound/tests/test_megno.py
+++ b/rebound/tests/test_megno.py
@@ -15,7 +15,7 @@ class TestMegno(unittest.TestCase):
         self.sim.add(m=1e-3,a=1.5,e=0.1,inc=0.1)
         self.sim.add(m=1.e-3, a=15., e=0.1, inc=0.1)
         self.sim.init_megno(seed=0)
-        self.sim.integrate(1000)
+        self.sim.integrate(10000.)
         self.assertAlmostEqual(self.sim.calculate_megno(),2.,delta=2e-1)
         self.assertAlmostEqual(self.sim.calculate_lyapunov(),0.,delta=1e-3)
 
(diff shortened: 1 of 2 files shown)


hannorein:
Thanks so much for catching this. Yes to 1). I don't think `keep_unysynchronized=1` and `safe_mode=1` make sense together (I guess that's why we haven't come across the issue until now). 

I guess it makes sense that `rand` is not machine independent, but I hadn't thought about it! I'm not sure what's best here. Implementing our own RNG sounds like a can of worms. When are random numbers needed for exact reproducibility? One approach would be to say that one can always restart the simulation form the first snapshot of the SimulationArchive (with all the random numbers for the initial conditions already drawn).

dtamayo:
I agree, it's definitely an edge case that most people won't run into. It was basically because I was trying to run lots of simulations run in different ways through the same function for the stability stuff. 

And I agree about the RNG and that the default should be to say that we only guarantee reproducibility when starting from a simulation_archive. That gets around all RNG problems and problems with initialization in python etc. I needed to seed the RNG for reproducibility for init_megno. The stability integrations were run without variational particles, but I want to add them for the short integrations. Of course the random init_megno better not impact the predictions, but I thought for consistency it would be nice to be able to reproducibly regnerate the exact same dataset from a script.

hannorein:
True, but if something depends on random numbers, one could argue that the user expects the script to return a different outcome every time it is run. And exactly the same results might lead to confusion. So maybe one should just make sure it's obvious in the documentation if a function call adds randomness.

dtamayo:
Makes sense. My naive hope was that I could seed the RNG and generate the same megno values on my laptop as my workstation. But it's not crucial. I agree the clearest requirement for reproducibility is that two people start from the same simulation archive. Thanks for talking it through!

PR #386, opened 2019-07-12
added Nsteps option to sim.step

dtamayo:
I want to get the times exactly right on a logarithmically spaced list of times. I think this is a useful option for these specialized cases...just lets you pass an integer number of steps to sim.step(). It follows the same convention as integrate in that Nsteps is the final number of steps you want to finish on (not how many steps to do). It reuses sim.steps_done to do that, so unless I'm missing logic where steps_done would not be correct, I think this should always work. I can now do things in a nice syntax like

```
times = np.logspace(0, np.log10(tmax), Nout)
Nsteps = np.int64(np.round(times/simWH.dt))
times = Nsteps*simWH.dt # corrected times

xWH, x = np.zeros(Nout), np.zeros(Nout)
for i, time in enumerate(times):
    sim.integrate(time)
    x[i] = sim.particles[1].x

for i, Nstep in enumerate(Nsteps):
    simWH.step(Nstep)
    xWH[i] = simWH.particles[1].x
```

### DIFF
--- a/rebound/simulation.py
+++ b/rebound/simulation.py
@@ -1498,12 +1498,19 @@ class Simulation(Structure):
         clibrebound.reb_output_binary(byref(self), c_char_p(filename.encode("ascii")))
 
 # Integration
-    def step(self):
+    def step(self, Nsteps=None):
         """
-        Perform exactly one integration step with REBOUND. This function is rarely needed.
+        Perform discrete integration steps with REBOUND. By default, executes a single step.
+        If Nsteps is passed, it will integrate to that TOTAL number of Nsteps.
+        Follows the same convention as integrate, e.g., if 3 steps have already been done
+        and you call sim.step(5), it will integrate until the total number of steps is 5
+        (not for 5 additional steps to 8). This function is rarely needed.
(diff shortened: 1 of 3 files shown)


hannorein:
Thanks! That looks very useful! I agree that it would be nice to use the same convention as for `integrate`, but I wonder if the opposite (passing it the number of steps it needs to take) would be more intuitive. Your code would still work if you changed the line to 

``
simWH.step(Nstep-simWH.steps_done)
``

Let me think about it for a minute!

dtamayo:
Good point, I agree you probably expect the other behavior. I guess if you want to integrate IAS to the same time you could skip all of this and do:

```
for i, time in enumerate(times):
    simWH.integrate(time, exact_finish_time=0)
    sim.integrate(simWH.steps_done*simWH.dt)
```

I'm happy to change it and add it in or leave it out altogether. Up to you

hannorein:
So I think it might be easiest to leave it out all together. It will avoid confusions and we have a workaround for our rather specific needs. Sound ok?

dtamayo:
Yup!

PR #387, opened 2019-07-20
reb_error instead of reb_exit

dtamayo:
```
sim = rebound.Simulation()
sim.simulationarchive_snapshot('path_doesnt_exist/file.bin')

```
would crash from python. Changed reb_exits to reb_errors in output.c. Also changed functions that might not get called from Python (not sure), but I didn't see a downside.


hannorein:
Thanks! That's a case I had not encountered yet!

dtamayo:
You need people doing weird things to find the edge cases ;)

PR #392, opened 2019-08-15
Newrebx

dtamayo:
So apart from removing REBOUNDx references  (and I got rid of the annoying callback warnings), I added an extra callback to reb_simulation that gets called when the  simulation is freed. I need to set that from REBOUNDx to know that the rebx->sim pointer is no longer valid in order not to seg fault.


hannorein:
Why is the `extras_cleanup` not in the `reb_simulation` struct in `rebound.h`?

hannorein:
We should also probably set `r->extras_cleanup` to `NULL` in `reb_reset_temporary_pointers()`.

PR #395, opened 2019-09-02
Create a .gitattributes file to fix language stats

dfm:
GitHub uses this file to decide which files to use and ignore when computing the language stats and this should correct for the fact that rebound is definitely not a `Jupyter Notebook` project. Feel free to ignore!


hannorein:
Thanks!

PR #398, opened 2019-11-05
fix compile errors with MPI=1

rometsch:
Compiling the **simplest** example code with MPI=1 didn't work for me. 

Here are two changes that fixed the issue for me:

The first change in Makefile.dev was necessary because in the examples/simplest/Makefile the variable **CC** was still set to **cc** instead of **mpicc** which caused an **mpi.h** not found error.

The second change is probably just some typo like error.

Let me know if this makes sense and if you have any comments

Cheers,
Thomas

For reference:
My system is Ubuntu 18.04  with the default openmpi packages installed via apt (libopenmpi-dev 2.1.1 and openmpi-bin 2.1.1).

### DIFF
--- a/src/Makefile.defs
+++ b/src/Makefile.defs
@@ -14,10 +14,10 @@ ifeq ($(OS), Darwin)
 endif
 
 ifeq ($(MPI), 1)
-	CC?=mpicc
+	CC=mpicc
 	PREDEF+= -DMPI
 else
-	CC?=cc
+	CC=cc
 endif
 
 ifeq ($(FFTW), 1)
--- a/src/Makefile.defs
+++ b/src/Makefile.defs
@@ -14,10 +14,14 @@ ifeq ($(OS), Darwin)
 endif
 
 ifeq ($(MPI), 1)
-	CC=mpicc
+	ifeq ($(origin CC),default)
+		CC:=mpicc
+	else
+		CC?=mpicc
+	endif
 	PREDEF+= -DMPI
 else
-	CC=cc
+	CC?=cc
 endif
 
 ifeq ($(FFTW), 1)
--- a/src/rebound.c
+++ b/src/rebound.c
@@ -732,11 +732,11 @@ static void* reb_integrate_raw(void* args){
     reb_sigint = 0;
     signal(SIGINT, reb_sigint_handler);
     struct reb_thread_info* thread_info = (struct reb_thread_info*)args;
+	struct reb_simulation* const r = thread_info->r;
 #ifdef MPI
     // Distribute particles
     reb_communication_mpi_distribute_particles(r);
 #endif // MPI
-    struct reb_simulation* const r = thread_info->r;
 
     double last_full_dt = r->dt; // need to store r->dt in case timestep gets artificially shrunk to meet exact_finish_time=1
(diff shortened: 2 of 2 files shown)


hannorein:
If you remove the question mark from `CC?=mpicc` you overwrite the `CC` environment variable if it's already set. I don't think this is a good idea. How you set the environment variable depends on your system, but try compiling the original version with `CC=mpicc make`.

rometsch:
Ok, so this is what the `?=` is for. I just observed what happens if I leave it out. Thanks for clarifying!

Coming back to the issue. I would assume that we want exactly the behavior you mentioned.
Running `MPI=1 make` should result in the use of `mpicc`, no matter what the environment variable `CC` is set to.

This is also how I read lines [16-21](https://github.com/hannorein/rebound/blob/850c85dd473f10bbf956f7515d9ab21d58137bdc/src/Makefile.defs#L16) in Makefile.defs


Please correct me if I misinterpret the `MPI=1` flag.

hannorein:
You system somewhere (`.bashrc`, another makefile such as in `examples/selfgravity_disc_mpi/Makefile`, etc) already sets the `CC` variable. You can try it out with `printenv CC` or a similar command. Because it's already set, the `?=` syntax does not overwrite it. This is on purpose. The original makefile lets you choose the compiler without editing the makefile by simply using `CC=mpicc make`. In your version, you would always have to edit the makefile to change the compiler.

rometsch:
For me the issue is, that I have no `CC` environment variable defined anywhere.
`printenv CC` returns nothing. Also `env | grep CC' doesn't return anything.

Here are some things I found out:

According to the [make manual on flavors](https://www.gnu.org/software/make/manual/make.html#Flavors) (scroll to the back of the section), the 
```
FOO ?= bar
```
assignment is is actually equivalent to
```
ifeq ($(origin FOO), undefined)
  FOO = bar
endif
 ```

The `origin FOO` is key here.  [origin's manual entry](https://www.gnu.org/software/make/manual/html_node/Origin-Function.html) describes that there are some variables which have a default value of `default` instead of `undefined` which then causes the `ifeq` to evaluate to false and the variable assignment via `?=` doesn't happen.

Turns out that `CC` is such a variable which is defined by an [implicit rule](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html#Implicit-Variables) and its default value is `cc`.
(You can run `make -p` in a directory that doesn't contain a makefile to see all implicitly defined variables.)

After fixing the line order in `rebound.c`, I can now compile the code with MPI enabled on a fresh copy of the rebound repo by running `MPI=1 make -R`.
The `-R` flag (equivalent to `--no-builtin-variables`) disables all implicitly defined variables.

To solve the issue without using the `-R` flag or manually setting `CC=mpicc ... make ...`, I pushed another commit.

`Makefile.defs` becomes a little more complex, but it restores the functionality I intuitively assumed for `MPI=1 make`.

hannorein:
I did not appreciate the complexity involved. Thanks a lot for looking into it. 

Looking at the MPI examples, it looks like I simply manually set both `MPI=1` and `CC=mpicc` (which makes the code in `Makefile.defs` redundant). 

Clearly your solution looks better!

rometsch:
Thanks a lot for your incredibly fast responses. They really kept me motivated digging all this up :)

PR #406, opened 2020-01-24
make REB_GRAVITY_NONE reset accelerations

dtamayo:
I *think* this is the right place to put this. REBOUNDx doesn't know when update_acceleration is being called (for a whole timestep, at a subtimestep), so I think REBOUND should be responsible for resetting the accelerations each timestep like it does in the other gravity routines, even in the case where it's set to NONE. I had this issue when making custom splitting schemes for the REBOUNDx paper where I added an operator post timestep to reset the accelerations, but this seems cleaner to me.


hannorein:
I think I agree. There might be a performance penalty. But it seems insignificant and  I can't think of any other negative side effects.

PR #419, opened 2020-03-30
mention that removing particles changes the references

### LINKED ISSUE #418: removing particles changes existing variables
Hi,

Thanks for the great project. I'm quite a beginner to n-body integration, but nevertheless rebound is easy to use.

This is not technical a bug, as it is working correctly, but not as one would expect and I think this could trip up beginners and might warrant a note in the docs.

Minimal example:
```python
from rebound import Simulation, Particle

sim = Simulation()
sim.units = ('yr', 'AU', 'kg')
sim.add(m=100, hash=0)
# let's add planets with easy to guess hashes
for i in range(1, 10):
    sim.add(a=i, hash=i, m=1e10)

# now let's pick two random particles (e.g. they were in a collision together like in CloseEncounters.ipynb)
cp1: Particle = sim.particles[3]
print(cp1.hash.value)  # returns '3' as expected
cp2: Particle = sim.particles[7]
print(cp2.hash.value)  # returns '7' as expected

# let's say we we want to remove those two particles from the simulation after the collision
print([p.hash.value for p in sim.particles])
sim.remove(hash=cp1.hash)
print([p.hash.value for p in sim.particles])  # now only [0, 1, 2, 4, 5, 6, 7, 8, 9] are left as 3 got removed
sim.remove(hash=cp2.hash)
print([p.hash.value for p in sim.particles])
# now one would expect 7 to also be gone
# but it seems like cp2 only points to the index in the particle list, so it was now in fact the particle with the hash 8
# as a result 8 is gone, but 7 remains
# [0, 1, 2, 4, 5, 6, 7, 9]
```
The solution is quite easy once one knows what happens:

```python
cp1_hash = cp1.hash
cp2_hash = cp2.hash

sim.remove(hash=cp1_hash)
sim.remove(hash=cp2_hash)
```

But finding out that this happened wasn't obvious until I saw the same particle collide multiple times.

Findus23:
fixes #418

### DIFF
--- a/ipython_examples/CloseEncounters.ipynb
+++ b/ipython_examples/CloseEncounters.ipynb
@@ -195,7 +195,7 @@
     "    sum_mass = cp1.m + cp2.m\n",
     "    mergedPlanet = (cp1*cp1.m + cp2*cp2.m)/sum_mass \n",
     "    mergedPlanet.m  = sum_mass\n",
-    "    sim.remove(index=col_i2) \n",
+    "    sim.remove(index=col_i2) # this changes the particles cp1 and cp2 point at \n",
     "    sim.remove(index=col_i1)\n",
     "    sim.add(mergedPlanet, assignHash=True)\n",
     "\n",


hannorein:
Thanks! But I'm not sure I understand the comment?

     this changes the particles cp1 and cp2 point at

Findus23:
Hi,

I tried to keep it short, but maybe the meaning got a bit lost.

I meant it in the sense of 
"Removing a particle in the simulation changes which particles the variables `cp1` and `cp2` point at. Printing `cp1.xyz` before and after this line might return different positions."

If you know a great way to word this, please let me know.

PR #421, opened 2020-04-03
Update Resonances_of_Jupiters_moons.ipynb

michaelaye:
add a newline so that readthedocs isn't messing things up.


hannorein:
Thanks!

PR #446, opened 2020-08-04
fix incorrect distance formula on 67P/Churyumov-Gerasimenko example

adalava:
This change fixes distance formula that was using X,Y instead of X,Y,Z. Also, increase Noutputs for better approximation


hannorein:
Thanks!

PR #447, opened 2020-08-20
Inconsistent archive behavior when deletefile=True

dtamayo:
I noticed when using the SPOCK Nbody class (which always overwrites archives with deleteFile=True) that when you do overwrite an existing archive, you no longer save the initial snapshot like you normally do when writing a new one. It seems useful to always save the initial condition and to have consistent behavior when a file does and does not exist.

I now reset all the auto intervals when deleting the file, which causes the C automate functions to update r->next the same way regardless of whether or not the file already exists. 

I don't see any side effects and I've checked this passes all tests on Travis, but I might be missing something!


hannorein:
Thanks for spotting that. I agree, this should be the expected behaviour.

PR #469, opened 2021-01-05
fix matplotlib deprecation warning

Findus23:
I got the following deprecation warning when using the latest version of matplotlib (3.3.3):

>rebound/rebound/plotting.py:83: MatplotlibDeprecationWarning: In a future version, 'pad' will default to rcParams['figure.subplot.wspace'].  Set pad=0 to keep the old behavior.
  ax_right = divider.append_axes("right", size="%.2f%%"%(100.*slices), sharey=ax_main)

As I think there shouldn't be any padding between the slices and the main plot, I just added the default of pad=0.


hannorein:
Thanks. As far as I can tell the pad parameter is old, so this shouldn't break anything.

PR #470, opened 2021-01-05
fix broken link to UniquelyIdentifyingParticles.ipynb

Findus23:
There are two more broken links to `Exceptions.html` and `Installation.html`, but I am not sure where they should point instead.


hannorein:
Thanks!

PR #471, opened 2021-01-05
update starman example

Findus23:
This one might have escalated a bit.
I just wanted to fix the broken URL in the beginning, but noticed that the website was updated to https://web.archive.org/web/20190524045516/https://www.projectpluto.com/temp/j95.htm later on which mentions that the original data was just an estimate from incomplete data which turned out to be incorrect and https://projectpluto.com/temp/spacex.htm#elements has the more correct data.

And I also found https://ssd.jpl.nasa.gov/horizons_batch.cgi?batch=1&COMMAND=-143205&CENTER=%27500@10%27&MAKE_EPHEM=YES&TABLE_TYPE=ELEMENTS&START_TIME=2018-05-01&STOP_TIME=%272018-05-01+00:00:01%27&OUT_UNITS=AU-D&REF_PLANE=ECLIPTIC&REF_SYSTEM=J2000&TP_TYPE=ABSOLUTE&ELEM_LABELS=YES&CSV_FORMAT=NO&OBJ_DATA=YES which has even newer data.

So I ran the notebook again with the other data and the results seem to be quite similar (but interestingly `c3 = 967`, but it might be that the original value was incorrect).


hannorein:
Thanks for looking into it. If the data is now in NASA Horizons, I guess it doesn't make much sense anymore to add it manually? We could just add it the same way as we add all other other objects. Or am I missing something?

Findus23:
Apart from the horizons link above, I can't find a way to query it to get the starman data as a response, but that might just be because I'm not really familiar with horizons (or it isn't a normal entry).

hannorein:
`sim.add("SpaceX Roadster")`

Findus23:
Okay, I didn't think it was that trivial :)
I even tested "Tesla Roadster" (the title of the Horizons page), but completely missed the "Target body name" field.
Interestingly we are now back at `c3 = 11.867790 (km^2/s^2)` and the OrbitPlot looks a bit more realistic

hannorein:
Great. I'll merge this in. 

Small note: the separate queries to HORIZONS work as long as one does not restart the ipython session. If no date is passed, then REBOUND uses the last date it used (to make sure the ephemeris are all consistent). It might in general be safer to just query NASA HORIZONS one, or add the date explicitly each time.

PR #472, opened 2021-01-05
add search field to sidebar

Findus23:
And one more while I'm on it:

I really like the sphinx search as it makes it really fast to find a specific page when you don't remember in which category it is left. But the overwritten sidebar template removes the default search field, so I readded it in this PR.
image

I can't promise that it works on ReadTheDocs as I think they use their own server-side search instead of the default sphinx search, but at least locally it works fine.


hannorein:
Seems to work! Thanks.

PR #474, opened 2021-01-06
fix typos in docs

Findus23:
I found some typos when reading through the docs, so I decided to quickly check all of them using [languagetool](https://languagetool.org/) (or rather [pylanguagetool](https://pylanguagetool.lw1.at/) which supports rst and ipynb files directly).

### DIFF
--- a/ipython_examples/AdvWHFast.ipynb
+++ b/ipython_examples/AdvWHFast.ipynb
@@ -16,7 +16,7 @@
     "As long as \n",
     "\n",
     "1. you don't add, remove or otherwise modify particles between timesteps\n",
-    "2. you get your outputs by passing a list of output times ahead of time and access the particles pointer between calls to `sim.integrate()` (see, e.g., the Visualization section of [WHFast.ipynb](WHFast.ipynb))\n",
+    "2. you get your outputs by passing a list of output times ahead of time and access the particle's pointer between calls to `sim.integrate()` (see, e.g., the Visualization section of [WHFast.ipynb](WHFast.ipynb))\n",
     "\n",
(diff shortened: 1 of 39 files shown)


ipython_examples/AdvWHFast.ipynb
```diff
@@ -16,7 +16,7 @@
     "As long as \n",
     "\n",
     "1. you don't add, remove or otherwise modify particles between timesteps\n",
-    "2. you get your outputs by passing a list of output times ahead of time and access the particles pointer between calls to `sim.integrate()` (see, e.g., the Visualization section of [WHFast.ipynb](WHFast.ipynb))\n",
+    "2. you get your outputs by passing a list of output times ahead of time and access the particle's pointer between calls to `sim.integrate()` (see, e.g., the Visualization section of [WHFast.ipynb](WHFast.ipynb))\n",
```
hannorein, line 19: I think that should remain particles, or ideally, `particles`.

hannorein (commented):
That is an embarrassingly large number of typos! Your corrections are much appreciated. Thank you!

rebound/simulationarchive.py
```diff
@@ -278,7 +278,7 @@ def getBezierPaths(self,origin=None):
         Examples
         --------
         The following example reads in a SimulationArchive and plots
-        the trajectories as Cubic Bezier Curves. It also plots the 
+        the trajectories as Cubic Bézier Curves. It also plots the
```
hannorein, line 281: Because the encoding is not explicitly specified, this fails the tests on travis. I'd suggest leaving it as Bezier.

PR #477, opened 2021-01-14
boxsize is not optional for configure_box()

Findus23:
todays minor PR is just something I noticed while reading through the code and I quickly wanted to edit before I forget it again


hannorein:
Thanks!

Issue #475, opened 2021-01-08
Issues with MERCURIUS and collision handling


Findus23:
Hello,

This time I have stumbled across a bit more complex issue, so this one will be a bit longer (I still tried to break it down to a minimal example).

A bit of background: I'm currently working on my Master's Thesis and my main plan would be to reproduce https://ui.adsabs.harvard.edu/abs/2020A%26A...634A..76B/abstract, but with a bit simplified collision handling (letting rebound do most of the work) and interpolation of a collision catalogue for water retention fractions instead of dedicated SPH simulations.
So the core part would be simulating late planetary formation and tracking water fractions over collisions. As this takes a long time, but also has a lot of collisions MERCURIUS seems like the obvious solution.

So this is the basis of the program that runs the simulation and on an encounter calculates collision speeds and angle 

```python
import numpy as np
from numpy import linalg
from rebound import Simulation, Collision

from data import add_particles

sim = Simulation()
sim.integrator = "mercurius"
sim.units = ('yr', 'AU', 'kg')
sim.dt = 0.005
sim.collision = "direct"
sim.ri_mercurius.hillfac = 3.

add_particles(sim)

t_max = 5000
t = 0
print(sim.G)
while t <= t_max:
    print(f"{t / t_max * 100:.2f}% done, N={sim.N}")
    try:
        sim.integrate(t)
        merc = sim.ri_mercurius
        print("dt", sim.dt)
        print("current mode", "ias15" if merc.mode else "whfast", merc.mode)
        t += 10
    except Collision:
        collided = []
        for p in sim.particles:
            if p.lastcollision >= sim.t - sim.dt:
                collided.append(p)
        assert len(collided) == 2
        p1, p2 = collided
        print(collided)
        # Custom resolution
        v1 = np.array(p1.vxyz)
        v2 = np.array(p2.vxyz)
        r1 = np.array(p1.xyz)
        r2 = np.array(p2.xyz)
        vdiff = v2 - v1
        rdiff = r2 - r1
        vdiff_n = linalg.norm(vdiff)
        rdiff_n = linalg.norm(rdiff)
        print("dt", sim.dt)
        merc = sim.ri_mercurius
        print(merc.mode)
        print("current mode", "ias15" if merc.mode else "whfast")

        print("rdiff", rdiff)
        print("vdiff", vdiff)
        print("sum_radii", p1.r + p2.r)
        print("rdiff_n", rdiff_n)
[...]

PR #483, opened 2021-01-29
Python Collision Resolve Example

Rmelikyan:
Hi Hanno,

I've seen a number of issues/questions come up regarding handling collisions and how to better define these interactions. This has been well documented for the C side of REBOUND, but for Python, a user defined collision resolve is just a small mention.

It took time for me to figure it out, but I eventually did implement my own function for my work and have prepared this example notebook that you might consider adding to the ipython examples for others to reference!

I tried to follow similar language and commentary as you provide in many of your examples, but of course will appreciate any edits or corrections you have for this.

I hope this can be a useful addition to this code base!

-Robert M.


Findus23:
Many thanks for the example, it nicely demonstrates the topic of #475.

Just a bit of feedback:
I think the `ps[i].x` lines are a bit verbose and hard to read. Maybe it is easier if you would before set the two particles to variables (`cp1=ps[i]`).

Also you don't need to add the three axis separately, something like `merged_planet = (cp1 * cp1.m + cp2 * cp2.m) / total_mass` should work directly.

I'd use ** instead of pow() as this is more familiar to non-C-users.

I had to use a diff-tool to find the difference between the two `my_merge()` as they look too similar. Maybe there is a way to merge them.

I'd also expand on the explanation why you would want to keep the fist particle or the more massive particle and what the implications of both are for things like N_active (in my Simulations I want the merged object to be of the same type as the more massive one, so I keep the more massive one and drop the other one)

s/vecotrs/vectors/g, s/reitterate/reiterate/g, s/impliment/implement :) 

A sentence in the beginning why this method might be better would be nice.


But this example should help people trying to simulate collisions a lot.

Rmelikyan:
Great comments! I'll apply much of your suggested simplifications for user clarity. I'll push that in soon.

hannorein:
Thanks to both of you. This is great. 

Just a bit of a general comment. This example is on the long side. Maybe it can be shortened without missing out on too much. The plots, for example, could be a bit simpler. I wouldn't bother putting all the label on, changing the font size, etc. That just distracts from the main point (how to write the collision_resolve function). 

I also had a hard time spotting the difference between the merge functions. Maybe one is enough? I don't think the second one really illustrates a new concept.

It's great that you added the doc string to your functions. But given this is a jupyter notebook, you could simply describe the function in a comment cell and don't bother with the doc string. (Again, just trying to keep it as concise as possible).

Rmelikyan:
Hi Hanno,

When you have a minute, check my most recent commit for an update to the example.

Let me know if there are other improvements to be made :)

-R

hannorein:
Thanks a lot! I've just shortened it a little more and corrected a few typos (and probably added a few more).

PR #489, opened 2021-02-04
Small typo fixes to new example

Rmelikyan:
Hi Hanno,

Reviewing the new example on the Rebound 'readthedocs' page I notice a few leftover errors. 

There is one issue I'm not certain I solved witch is the clipping of the first 3-4 words of the markdown text after the title. I added an extra line spacing that may be the fix, but I've seen this same bug on a couple other of the examples when presented in the online format so I wanted to make note of that here.

Attaching screenshot examples of intro sentence clipping from rebound.readthedocs

Screen Shot 2021-02-04 at 1 55 54 PM

Screen Shot 2021-02-04 at 1 57 16 PM

Screen Shot 2021-02-04 at 1 58 02 PM

Screen Shot 2021-02-04 at 1 58 20 PM

Notice how a full sentence is missing from the top of this example... I'm not experienced at all with readthedocs, but I hope that documenting the problem will bring insight who is experienced!

Best
-Robert

PR #490, opened 2021-02-05
fix iPython docs

Findus23:
I really don't understand the logic behind `emptylines`, but if the point of this code is just to add ` (iPython)` to all iPython headings, then I think this can be achieved a lot simpler.
Also this doesn't add the unneeded `==========` as described in https://github.com/hannorein/rebound/pull/489#issuecomment-773988482

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -12,11 +12,9 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys
-import os
-import shlex
-import subprocess
 import glob
+import os
+import subprocess
 
 # Doxygen trigger
 read_the_docs_build = os.environ.get('READTHEDOCS', None) == 'True'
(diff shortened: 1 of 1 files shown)


hannorein:
I don't remember the details. There was as bug in some module (readthedocs, nbconvert?) which either added an empty line or needed an empty line to correctly render the examples. Or the file wasn't included in the index. Something like that. 

Did you check if this fix affects any other examples (c or python)? Does the index still work?

Findus23:
I can't speak for readthedocs, but locally both the ipython and the C examples look like before and if by index you are referring to `/examples.html` then this also look like before. 

To be sure I now also deleted the doc/ipython and doc/build directories and it still works

hannorein:
Ok! I'll merge it, then let's see if readthedocs does something funny with it. 

Note to past-self: write better commit messages.

Findus23:
In the meantime I also did a diff of the build between main and this branch (not to past-self: check spelling before creating branches) and I can't find any changes apart from the missing first lines.

Findus23:
seems to be looking good: https://rebound.readthedocs.io/en/latest/ipython/User_Defined_Collision_Resolve.html

hannorein:
Yep. Thanks for looking into this!

Rmelikyan:
Wow! Nice fix to the docs!

Issue #491, opened 2021-02-08
C heartbeat in python doesn't work


Findus23:
I honestly haven't looked into that topic too far, but I even fail to just run the example jupyter notebook without any custom modifications:
https://rebound.readthedocs.io/en/latest/ipython/Cheartbeat.html

Running the notebook (after running `make` in the main rebound directory) works fine up until `gcc -c -O3 heartbeat.c -o heartbeat.o` which also works, but `gcc -L. -shared heartbeat.o -o heartbeat.so -lrebound` then fails with an error
```text
/usr/bin/ld: heartbeat.o: warning: relocation against `e' in read-only section `.text'
/usr/bin/ld: heartbeat.o: relocation R_X86_64_PC32 against symbol `e' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
```
interestingly l can see `-fPIC` in the librebound Makefile, so I'm not entirely sure what goes wrong.
https://github.com/hannorein/rebound/blob/c55f9de3401815f2d5b3647e21deda4e26de67be/src/Makefile#L3

About my system
OS: Debian Testing
gcc Version: 10.2.1
```bash
➜  ~ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/10/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa:hsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
[...]

PR #492, opened 2021-02-09
extend C heartbeat example

Findus23:
extend the example with hints from https://github.com/hannorein/rebound/issues/491


hannorein:
Thanks! Still a hack, but it might help people having the same issues.

PR #493, opened 2021-02-09
Ipython notebook check

Findus23:
To avoid an issue like https://github.com/hannorein/rebound/issues/491 where I can't reproduce the output of an example notebook in the future, I quickly tried running all of them to check if the output is still similar to the stored one. (and all of them worked)

I updated the code in OrbitalElements.ipynb and TransitTimingVariations.ipynb to avoid deprecation warnings and reran Testparticles.ipynb and Units.ipynb as their output was from a really old rebound version.


ipython_examples/TransitTimingVariations.ipynb
```diff
@@ -101,7 +101,7 @@
    "outputs": [],
    "source": [
     "A = np.vstack([np.ones(N), range(N)]).T\n",
-    "c, m = np.linalg.lstsq(A, transittimes)[0]"
+    "c, m = np.linalg.lstsq(A, transittimes, rcond=None)[0]"
```
Findus23, line 104: Seems like the tests use a version that doesn't support this yet: https://travis-ci.com/github/Findus23/rebound/jobs/481686474#L897-L903

hannorein: What was the issue without `rcond`?

hannorein: Just the warning message? Or was there an issue with the result?

Findus23: The result is fine.
It's just that numpy is warning:
```
<ipython-input-4-5c626cb5e957>:2: FutureWarning: `rcond` parameter will change to the default of machine precision times ``max(M, N)`` where M and N are the input matrix dimensions.
To use the future default and silence this warning we advise to pass `rcond=None`, to keep using the old, explicitly pass `rcond=-1`.
  c, m = np.linalg.lstsq(A, transittimes)[0]
```
https://numpy.org/doc/stable/reference/generated/numpy.linalg.lstsq.html

hannorein: But the tests fail...

Findus23: I guess the tests use a numpy version where None wasn't yet a valid parameter.
I switched it now to rcond=-1 which should keep the old behaviour and should also be a valid parameter in the old version.

hannorein:
Ok. Thanks. Just waiting for the tests to complete.

hannorein:
Finally!

PR #494, opened 2021-02-10
modification to mercurius so massless test-particle encounters don't affect planet evolution

katvolk:
A common use-case for hybrid integrators is integrating large sets of massless test particles with a system of massive bodies. It is useful for the evolution of the planets starting from a single set of initial conditions to remain identical even when different sets of test particles are evolved (for example, when you want to integrate thousands of test particles by spitting them across many rebound instances on a cluster). However, the fact that the test particles can trigger the close-encounter routine means that they do change the planet evolution even when they are massless (example plot of changes in Jupiter's mean anomaly attached). This can cause small problems (e.g., the planet phases between different instances are different, so you have to be careful when calculating phase-dependent properties like resonant angles to not combine data across different runs) or large problems (e.g., you're trying to study how planet migration affects the Kuiper belt using artificially induced migration plus test particles; a set of initial conditions and migration parameters that ends with acceptable giant planet orbits with one set of test particles could end with ejected ice giants for a different set of particles because a small phase-shift led to an unfavorable resonance crossing between the giant planets).

I have made some small changes to mercurius to prevent the test particles from affecting the planets by reverting the planet states at the end of a close-encounter step to the already-calculated post-kepler step states if there are no close encounters between bodies with mass (if the test particles have mass, or the encounters involve two planets, then the planet evolution from the close-encounter routine is kept!). This fixes the problems described above. (Though I have not yet tested this implementation with additional forces to make sure it doesn't change how those behave).

delta-ma

### DIFF
--- a/src/integrator_mercurius.c
+++ b/src/integrator_mercurius.c
@@ -159,7 +159,8 @@ static void reb_mercurius_encounter_predict(struct reb_simulation* const r){
     const double dt = r->dt;
     rim->encounterN = 1;
     rim->encounter_map[0] = 1;
-    for (int i=1; i<N; i++){
+    rim->tponly_encounter = 1;
+	 for (int i=1; i<N; i++){
         rim->encounter_map[i] = 0;
     }
     for (int i=0; i<N_active; i++){
@@ -216,6 +217,9 @@ static void reb_mercurius_encounter_predict(struct reb_simulation* const r){
                     rim->encounter_map[j] = j;
                     rim->encounterN++;
                 }
+                if (j<N_active || r->testparticle_type==1){
+                    rim->tponly_encounter = 0;
+                }					 
             }
         }
     }
@@ -342,6 +346,14 @@ static void reb_mercurius_encounter_step(struct reb_simulation* const r, const d
         }
     }
 
+    // if only test particles encountered massive bodies, reset the
+    // massive body coordinates to their post Kepler step state
+    if(rim->tponly_encounter){
+        for (int i=1;i<r->N_active;i++){
+            r->particles[i] = rim->particles_post[i];
+        }
+    }
+
     // Reset constant for global particles
     r->t = old_t;
     r->dt = old_dt;
@@ -399,6 +411,7 @@ void reb_integrator_mercurius_part1(struct reb_simulation* r){
         // These arrays are only used within one timestep. 
         // Can be recreated without loosing bit-wise reproducibility
         rim->particles_backup   = realloc(rim->particles_backup,sizeof(struct reb_particle)*N);
+        rim->particles_post   = realloc(rim->particles_post,sizeof(struct reb_particle)*N);
         rim->encounter_map      = realloc(rim->encounter_map,sizeof(int)*N);
         rim->allocatedN = N;
     }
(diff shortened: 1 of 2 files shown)


hannorein:
Hi Kat! That's awesome. You raise a good point which I had not thought about before. I think you've done the right thing when you implemented this an an optional flag. It might not be the default behaviour that one naively expects in most circumstances. 

Let me think a little more about the implications a little more before merging this in. There are also a few more bookkeeping items to make this work (I can do those, just putting them here for reference):

- [ ] Add the new variable to the input/output routines so that it gets stored in the binary files. 
- [ ] Add the variable to the python interface (right now it will crash).

katvolk:
Thanks Hanno! I did consider whether there should be some higher-level flag the user could set to either trigger this option or not, but I wasn't sure how best to do that. Right now it will always behave the way I want it to for test particles (I need it to for a project I'm starting!). I know there are other people who do high-throughput computing like I do that would appreciate this option, but you would have a better idea about whether this is the best way to implement it!

danielk333:
Just to confirm this usefulness to other users: I actually (wrongly now I realize) assumed that this WAS the default behavior as test-particles with no mass by definition has no feedback with the system dynamics. I usually parallelize all my code with MPI (so I can run it on home-built clusters/my laptop/HPC without any change) and this would have given my a nasty surprise. Thanks for the heads up!

hannorein:
It can't be the default behaviour because massive particles could never have close encounters with other massive particles. Even for encounters with massless particles, a user might want to manually modify a massive particle during a collision/encounter with a test particles. That would be overwritten with this new flag turned on. Just for reference, this only affects the dynamics of planets because of round-off errors (which can be important in a chaotic system as Kat has shown).

katvolk:
Just to be clear. The current modification does keep the close encounter calculations if they were two massive bodies involved in a close encounter. It only replaces the planets post close encounter state with the post Kepler step state if there were only test particles involved in the close encounter. That’s what the flag is keeping track of.

hannorein:
That leads to an issue where there are two planets and each is having close encounters with different test particles at the same time, but not with each other.

katvolk:
The flag gets set in the close encounter prediction routine, and I think that is accounted for.

hannorein:
Hm. Maybe I'm missing something, but I don't think so. If you just add another planet to your simulation, and that planet also has close encounters with test particles, then the positions of the planets will not get reset and you end up with the same issue you were trying to solve with the flag in the first place. 

If you're just interested in one planet plus test particles, then that's not a problem.

hannorein:
Never mind. I'm wrong. You're right. I understand now. Sorry! 

Let me think through the implications a little more. There are so many edge cases to consider...

katvolk:
Whew, I was just double checking to make sure is implemented the flag reset where I thought I had! Definitely worth thinking for a while about edge cases before implementing. I am almost always doing solar system cases, where this is fine, but there could be other cases I haven’t thought hard about where it leads to undesired changes. In which case it might be better to make this behavior optional with another user-set flag parameter.

hannorein:
I've pushed some changes to the katvolk-main branch on my repository. I'm now reusing the `particles_backup` array. This allows us to get away without creating a new `particles_post` array. It works because `particles_backup` is no longer used after setting up the particles for the encounter step. I've also added a unit test and fixed the python structures.  

Still thinking about possible unintended side effects...

PR #498, opened 2021-02-15
add missing import to __init__.py

Findus23:
Maybe this is again something where I am overthinking a trivial fact, but the `__all__` array lists all imports, but `reb_simulation_integrator_ias15` isn't imported above.

Also this is weird:

https://github.com/hannorein/rebound/blob/ed163305cc1dd9e314e80d686a564498d1f35ff1/rebound/__init__.py#L35-L42
`pkg_resources` is being used without being imported, which could never work.

### DIFF
--- a/rebound/__init__.py
+++ b/rebound/__init__.py
@@ -41,8 +41,8 @@ except:
     # Might fails on python3 versions, but not important
     pass
 
-# Exceptions    
-class SimulationError(Exception):  
+# Exceptions
+class SimulationError(Exception):
     """The simulation exited with a generic error."""
     pass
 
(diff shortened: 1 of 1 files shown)


hannorein:
I'm not sure the ias15 structure is really needed in `__all__`. But you're right, to be consistent with the other structures it would make sense to add it.

I guess at some point we must have removed `import pkg_resources` and because of the try catch block, we never saw the error message. It looks like `pkg_resources` is imported in `simulation.py`. But it's actually not needed there. Just moving the import statement to `__init__.py` seems to work for me. I can't think of any obvious problem doing that (we could add the import statement to the try catch block to be absolutely sure we avoid any issues).

Given it's 2021, I'm a bit reluctant to invest a lot of time thinking about python2.

Findus23:
Indeed it seems like `from rebound import reb_simulation_integrator_ias15` works even now, but for consistency sake it shouldn't hurt to add it.

Weirdly there was never a `import pkg_resources` that got removed. (https://github.com/hannorein/rebound/commit/1feb3bcbdf90bf55e242e51a9740b4c30f9ff8d3 introduced the code as-is)

https://github.com/hannorein/rebound/blob/6597aed52b8c48fc70a70514762f9993964f47b1/rebound/simulation.py#L19-L23

At least for me `python3.9 -c "import pkg_resources"` works fine, but of course I can't promise that this is the case for all python setups (I think it depends on the setuptools version).

hannorein:
Ah, well that commit explains it. The code was in `simulation.py` at some point, then got copied to `__init__.py` but without the import statements. This never popped up as an error because of the try catch block. So moving the import statement over seems to be the solution.

PR #500, opened 2021-02-15
run tests on Python 3.9

Findus23:
While rebound works with Python 3.9 it would be great to verify this by running the tests with 3.9 instead of 3.8 (hoping there won't be any 3.5-3.8 exclusive bugs in the future)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,18 +13,18 @@ env:
 python:
   - "2.7"
   - "3.4"
-  - "3.8"
+  - "3.9"
 
 matrix:
   exclude:
     - env: TESTPART=C
       python: "3.4"
     - env: TESTPART=C
-      python: "3.8"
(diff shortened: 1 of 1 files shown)


Findus23:
BTW: travis-ci.org is shutting down very soon and all projects have to migrate to travis-ci.com (which is probably why the test are insanely slow at the moment)

hannorein:
Yeah, I saw the notice, but last time I tried to migrate it, it didn't work for some reason and the message still says "Please stay tuned here for more information."

(I'm also not sure if travis is the CI we want to continue using in the future.)

Findus23:
Indeed until recently it said that it will be shut down in December, but it seems to have changed. 
Thankfully the tests don't need any odd dependencies or services, so they should run fine just about everywhere.

PR #511, opened 2021-02-26
various code style fixes

Findus23:
a few things that don't warrant a PR each:
- comparisons with None should always be done with `is None` (https://www.python.org/dev/peps/pep-0008/#programming-recommendations)
- a few missing parameters in the docstrings
- removing some trailing semicolons from C muscle memory :)
- `super(Class, self)` is supposed to be used with the superclass inside the `super()` call.

### DIFF
--- a/rebound/interruptible_pool.py
+++ b/rebound/interruptible_pool.py
@@ -70,7 +70,7 @@ class InterruptiblePool(Pool):
     def __init__(self, processes=None, initializer=None, initargs=(),
                  **kwargs):
         new_initializer = functools.partial(_initializer_wrapper, initializer)
-        super(Pool, self).__init__(processes, new_initializer,
+        super(InterruptiblePool, self).__init__(processes, new_initializer,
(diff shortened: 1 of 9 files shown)


hannorein:
Thanks. Are you sure about the super() syntax. I think I understand the [documentation](https://docs.python.org/3/library/functions.html#super) differently thank you. The one example they list is this:
```
class C(B):
    def method(self, arg):
        super().method(arg)   
```
The comment says the last line is the same as calling  `super(C, self).method(arg)`. Isn't that exactly what was in REBOUND before?

hannorein:
Looks like travis agrees with me 😉

Findus23:
You are of course right. I got myself completely confused.

hannorein:
You almost convinced me! 

Thanks for the other changes. They all look good (but I'll way for travis to be sure). When I find some time, I'll look into linting a bit more. There are quite a few other inconsistencies I'd like to clean up. And as always, your help is much appreciated!

PR #515, opened 2021-03-02
replace http links with https

Findus23:
nowadays there shouldn't really be any reason to not use HTTPS.
So I replaced all links apart from `http://www.gnu.org/licenses/` (as it is more of a specifier than a link) and `adsabs.harvard.edu` (as for whatever weird reason that domain has no HTTPS server doing a redirect to the current site)

Other changes:
- removed `--no-check-certificate` from wget
- replace mercury download with archive.org link (the original file seems to be offline since a few years ago)


hannorein:
Hm. I'm not sure about this one. I guess most of the changes seem unproblematic. The `--no-check-certificate` is there because a lot of operating systems have outdated certificates installed. Yes, one should update outdated certificates, but we really don't have to worry about any security risks for this type of package.

Findus23:
Hm, I honestly thought the time when the os ca-certificates where so outdated that one could not use major websites like GitHub were over (how do these people use any website nowadays?).
Then again you are of course right that for scientific software security isn't that important. I am just used to software that is used by many people and where it is not impossible that a targeted attacker would try to intercept requests until they find one that is not secured and where the sent data is executed. 

I guess keeping it in for people who don't read, but would complain about broken downloads, but adding an information sentence below it for the people who do read might be the better solution.

hannorein:
I'm probably overthinking it, but imagine someone who has set up a compute node somewhere in their office 10 years ago, or use some outdated department cluster where they don't have root access. If they just ssh into it and run some scientific software, there are no real security issues. And updating the whole operating system might be too big of an undertaking. People don't want to do that just to install an N-body integrator and they might go with the path of least resistance (i.e. copy and pasting mercury from the directory where they've used it before).

PR #535, opened 2021-05-10
spellcheck new docs

Findus23:
I didn't read through all of them yet, but just checked them using [pyLanguagetool](https://pylanguagetool.lw1.at/).


hannorein:
Amazing. Thanks!

PR #554, opened 2021-07-07
Corrupt

dtamayo:
A possible failure mode when interrupting a large number of simulations storing simulation archives is that a small subset might interrupt while writing a binary snapshot, corrupting them. The latest version allows you to read in such corrupted simulation archives but can't resume and continue appending from an archive that's corrupted in this way. This pull request changes things so that instead of appending to the end of the file, it appends snapshots to the last valid snapshot. 

Thanks for your help @hannorein . Took a little longer than I had hoped to figure out that when I truncated a valid archive to make a corrupted archive test case, the rest of the archive still stayed in contiguous memory so the code would read it anyway  😅

### DIFF
--- a/src/simulationarchive.c
+++ b/src/simulationarchive.c
@@ -609,6 +609,10 @@ void reb_simulationarchive_snapshot(struct reb_simulation* const r, const char* 
                 fseek(of, field.size, SEEK_CUR);
             }while(field.type!=REB_BINARY_FIELD_TYPE_END && bytesread);
             long size_old = ftell(of);
+            
+            // store location after initial binary (including trailing blob) as first valid end of file
+            long last_blob=ftell(of) + sizeof(struct reb_simulationarchive_blob);
+
             char* buf_old = malloc(size_old);
             fseek(of, 0, SEEK_SET);  
             fread(buf_old, size_old,1,of);
@@ -622,13 +626,31 @@ void reb_simulationarchive_snapshot(struct reb_simulation* const r, const char* 
             char* buf_diff;
             size_t size_diff;
             reb_binary_diff(buf_old, size_old, buf_new, size_new, &buf_diff, &size_diff);
-            
-            // Update blob info and Write diff to binary file
+    
             struct reb_simulationarchive_blob blob = {0};
-            fseek(of, -sizeof(struct reb_simulationarchive_blob), SEEK_END);  
+            fseek(of, size_old, SEEK_SET);
+         
+            // Find last valid snapshot to allow for restarting and appending to archives where last snapshot was cut off
+            do
+            {
+                fseek(of, -sizeof(struct reb_binary_field), SEEK_CUR);
+                fread(&field, sizeof(struct reb_binary_field), 1, of);
+                if (field.type != REB_BINARY_FIELD_TYPE_END){ // could be EOF or corrupt snapshot
+                    break;
+                }
+                bytesread = fread(&blob, sizeof(struct reb_simulationarchive_blob), 1, of);
+                last_blob = ftell(of);
+                fseek(of, blob.offset_next, SEEK_CUR);
+            } while(bytesread);
+
+            // To append diff, seek to last valid location (=EOF if all snapshots valid)
+            fseek(of, last_blob, SEEK_SET);
+
+            // Update blob info and Write diff to binary file
+            fseek(of, -sizeof(struct reb_simulationarchive_blob), SEEK_CUR);  
             fread(&blob, sizeof(struct reb_simulationarchive_blob), 1, of);
             blob.offset_next = size_diff+sizeof(struct reb_binary_field);
-            fseek(of, -sizeof(struct reb_simulationarchive_blob), SEEK_END);  
+            fseek(of, -sizeof(struct reb_simulationarchive_blob), SEEK_CUR);  
             fwrite(&blob, sizeof(struct reb_simulationarchive_blob), 1, of);
             fwrite(buf_diff, size_diff, 1, of); 
             field.type = REB_BINARY_FIELD_TYPE_END;
--- a/rebound/tests/test_simulationarchive.py
+++ b/rebound/tests/test_simulationarchive.py
@@ -546,7 +546,7 @@ class TestSimulationArchiveMercurius(unittest.TestCase):
         sim.integrate(3001)
         with open('simulationarchive.bin', 'r+b') as f:
             f.seek(15900)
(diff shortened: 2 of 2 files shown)


src/simulationarchive.c
```diff
@@ -609,6 +609,10 @@ void reb_simulationarchive_snapshot(struct reb_simulation* const r, const char*
                 fseek(of, field.size, SEEK_CUR);
             }while(field.type!=REB_BINARY_FIELD_TYPE_END && bytesread);
             long size_old = ftell(of);
+            
+            // store location after initial binary (including trailing blob) as first valid end of file
+            long last_blob=ftell(of) + sizeof(struct reb_simulationarchive_blob);
```
hannorein, line 614: This variable seems to get overwritten before its value is ever read.

src/simulationarchive.c
```diff
@@ -622,13 +626,31 @@ void reb_simulationarchive_snapshot(struct reb_simulation* const r, const char*
...
-            
-            // Update blob info and Write diff to binary file
+    
             struct reb_simulationarchive_blob blob = {0};
-            fseek(of, -sizeof(struct reb_simulationarchive_blob), SEEK_END);  
+            fseek(of, size_old, SEEK_SET);
+         
+            // Find last valid snapshot to allow for restarting and appending to archives where last snapshot was cut off
+            do
+            {
```
hannorein, line 635: I think this loop is problematic. We should probably not entered it by default. Only for corrupt archives. It could be very slow to scan through a few GBs of data everytime a snapshot is appended. Note that in the current implementation we only read in the beginning and the end.

hannorein (commented):
Thanks for this! I worry a bit about the performance with the new loop. I'm not entirely sure how to check if the file is corrupt without scanning through the full file. I guess it would be ideal to add a checksum-type parameter to `reb_simulationarchive_blob`, maybe containing the current location in the file. Then we could read in the last `sizeof(reb_simulationarchive_blob)` bytes from a file and check if the location matches. If not, the file is corrupt. But that would change the format of the binary files. Alternatively, we could check for the last `reb_simulationarchive_blob` to have a size of 0 and the last `reb_binary_field` to have an `offset_next` of 0 and a `size` of zero. But checking for zeros might give us false positives in many cases (because many arrays may contain a lot of zeros).

src/simulationarchive.c
dtamayo: Good catch! This was left over from when I wrote the do while loop as a while loop...can definitely be removed!

dtamayo:
Thanks @hannorein. Is it that slow to seek and read a couple fields Nblobs times? Do you have a GB binary I could get some intuition with? 

Naively I might have thought that if you are structuring the number of snapshots you store to avoid having to do excessive recomputation (say you don't want to redo more than 1 min of computation for an integration that takes 1 week total, that's ~10,000 snapshots). Is traversing 10,000 blobs comparable to 1 min?

dtamayo:
But assuming it is too slow (and by now I should know better than to question your intuition!), I agree that checking for zero could cause lots of problems, the checksum solution does seem ideal at the expense of changing the structure!

hannorein:
In the past, I have noticed  (to be more precise @zyrxvo has noticed) that it take be several seconds to scan through a SA on a local disk. That's why I've implemented the `reuse_index` option which can speed up loading SAs significantly.  But It strongly depends on the type of storage.

I'm not sure what the right solution is. I'll think about it a bit longer!

hannorein:
I pushed an additional commit to my [dtamayo-corrupt branch](https://github.com/hannorein/rebound/tree/dtamayo-corrupt).

- I think the original unit test you wrote passed by accident (because everything was overwritten with zeros). If you look closely, you see warning messages about the simulation not being synchronized. That warning only showed up because something went wrong when reading in the binary data. (It's an IAS15 simulation, so this message doesn't make sense)
- I added a new test to see if the file is corrupt by checking if the last two simulationarchive_blobs in the archive have consistent offset_next and offset_prev values. If not, something went wrong, like a checksum! Then I fall back to your loop, starting from the beginning of the file. This should be much faster than always starting from the beginning for every snapshot. 
- I added one more unit test which cuts the file short by a few bytes. I think this might be a more realistic failure scenario (i.e. when disk space is full) rather than overwriting the file with zeros. 
- If you have time, please check if this works on your end. There are a lot of changes to various logic statements which I could easily get wrong!

Also, I really need to add some documentation about the binary format!

dtamayo:
Thank you for the lightning fast patch Hanno..the checksum workaround is really clever!

I'm a bit confused about my unit test. In my pull request version, the following code does not give me any synchronization warnings when I load up different snapshots, and I can take those reloaded simulations and integrate them without errors

```
import rebound

sim = rebound.Simulation()
sim.add(m=1.)
sim.add(m=1e-3,a=1.)
sim.add(m=5e-3,a=2.25)

sim.automateSimulationArchive("simulationarchive.bin", interval=1000,deletefile=True) 
sim.integrate(3001)
with open('simulationarchive.bin', 'r+b') as f:
    f.seek(15900)
    f.write(bytes(72))
sa = rebound.SimulationArchive("simulationarchive.bin")
sim = sa[-1]
sim.automateSimulationArchive("simulationarchive.bin", interval=1000)
sim.integrate(7001)
sa = rebound.SimulationArchive("simulationarchive.bin")
```

With the dtamayo-corrupt branch, if I run the same code, then I do get synchronization warnings when I load the first appended snapshot `sim = sa[3]`, and I get a malloc error and the kernel dies if I try to load any snapshots after that. I also get the same behavior when I use your better code avoiding my magic numbers using os.SEEK_SET etc. in your updated unit test. 

As for the f.truncate() unit test, doesn't that run into the problem that the rest of the snapshot remains in contiguous memory? I haven't tested it carefully, but I would think that the checksum test would fail starting from SEEK_END, but that when you fall back to my loop, it would happily read past the end of the file and say that the truncated snapshot was still valid. At least that was the behavior I kept running into, but perhaps your logic catches this!

hannorein:
Hm. Thanks! I'll look into it.

hannorein:
Very strange. This works for me. What python version are you using? (me: Python 3.8.2)

hannorein:
Never mind. I do get the error if I load `sa[3]`. I think the problem is that "your" truncated file never shows up as corrupt.. Looking into it ... ;-)

dtamayo:
Thanks! I'm using python 3.7.7

hannorein:
Next try. I think this should work. Your test case with the zeros was a good one (and driving me nuts) ;-)

dtamayo:
This works great for me! Thank you so much for working on this so quickly...I assure you it took me much longer than it would have taken you to sort out a good test case and it was definitely driving me nuts too 😃

PR #566, opened 2021-09-03
correct names of integrators in example code lines

rcdorsey:
Mis-matching integrator names between C and Python code examples, as well as incorrect integrator names in the code under the different integrator subsections


hannorein:
Good catch! Much appreciated!

PR #573, opened 2021-09-30
Fixing the dual code displays

rcdorsey:
Not sure how to explain the problem but here goes. On most pages, any inline code is displayed for C and python in the same box, you just click "C" or "python" to view one or the other. On this 'Simulation Archive' page, the code for C and python were being displayed separately, and I think it was because there was a missing "=" before the "C" condition? That's my basic understanding at least.


hannorein:
Thanks for spotting this!

PR #587, opened 2022-02-11
TES

hannorein:
@peterbartram's TES simplementation


hannorein:
Todo list:

- [ ] Documentation.
- [ ] Show example where TES is better/faster than other integrators.
- [ ] Ideally show a few cases where *not* to use TES.

hannorein:
Given the unit tests that use `reb_diff_simulations` fail, I suspect there is an issue where some TES variables are not set to the right default value.

PR #588, opened 2022-02-11
new poincare surface of section example

shadden:
I've added a short new example that creates a poincare surface of section from integrations of the restricted three-body problem.


hannorein:
Sorry that it took me a while. It's now in the main branch. I've cleanup it up just a little, including
- I removed some of the plot settings
- using `import rebound` rather than `import rebound as rb`
I'm trying to keep these examples as minimalistic as possible in the hope this helps people just starting out.

PR #593, opened 2022-04-24
fix(readme): readme visualization in pypi

Baelfire18:
Currently, in the [PyPI web page of this library](https://pypi.org/project/rebound/) you cannot visualize the description the way you are supposed to. That is because the default readme format in PyPI is .txt, meanwhile in `rebound` the readme is a markdown file.
So with this new line, your next release would look as awesome in PyPI as [the official documentation web page](https://pypi.org/project/rebound/).


Baelfire18:
Here it is the official documentation that checks it 
https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/

hannorein:
Thanks! I had not noticed that before!

PR #639, opened 2023-01-03
added edge case test to to_new_axes and added failing unit test

dtamayo:
Not really a patch...I added a failing unit test for an edge case in from_to that I think will come up a lot (when trying to rotate to the invariable plane when you're already aligned with it, so the line of nodes is undefined).

`rebound.Rotation.from_to([1,0,0], [-1,0,0])`

yields

<rebound.simulation.Rotation object at 0x1057a9240, ix=nan, iy=nan, iz=nan, r=nan>

I also added testing for that edge case in Rotation.to_new_axes in python


hannorein:
Should be fixed now.

PR #642, opened 2023-01-10
Initialize webgl properly for newly loaded simulations

alchzh:
When using `takeScreenshot` to create a movie in Jupyter, the archive loading crashes the kernel. Probably caused by the new behavior of SimulationArchive.getSimulation allocating a new simulation object from scratch on call. This PR fixes it by calling `reb_display_init_data` on each load.


hannorein:
Thanks for catching and fixing that!

PR #646, opened 2023-01-13
added add sub mul div

dtamayo:
This is great!

I've added __add__ etc. so that we can write e.g.

Ltot = sim.angular_momentum() + rebx.spin_angular_momentum()

Should we make __repr__ print something less ctypesy, closer to Vec3d([1,2,3])?

I added a little bit of text to the orbit classmethod for Rotation, which I see has given up on putting a 'from' or a 'to' 😂


hannorein:
Thanks!

Yes to your question.

PR #647, opened 2023-01-13
acos2 is now static

hannorein:
@dtamayo I noticed that the acos2 function did not have the `reb_` prefix but was part of the shared library. That could lead to issues when there is another library implementing the same function. It looks like we never use the function outside of `tools.c` so I have made it static. It looks like you also don't use it in reboundx. But maybe I'm missing something. Asking you because you implemented it. If not, I'll merge this in.

### DIFF
--- a/src/tools.c
+++ b/src/tools.c
@@ -933,7 +933,10 @@ struct reb_orbit reb_orbit_nan(void){
 // will return 0 or pi appropriately if num is larger than denom by machine precision
 // and will return 0 if denom is exactly 0.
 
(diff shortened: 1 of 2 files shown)


dtamayo:
Yup, this can be static! I needed it for reb_vec3d so I changed things when they were in REBOUNDx, but not anymore. Thanks for catching it. There's no rush on this, but could we make a new REBOUND version once the vec3d stuff is settled? I need the new REBOUND version to pass the REBOUNDx unit tests. Thanks again!

hannorein:
done

PR #649, opened 2023-01-17
Properly free tree cells in reb_tree_delete_cell

alchzh:
I think there was a missing call to `free()` here (the function does... nothing at the moment)

Fixes a small memory leak reported in valgrind.


hannorein:
Ha. Thanks for that. I believe you're right. 

I think this function only gets called if one deletes the entire tree (typically only at the end of a simulation). I don't think it affects the code while it's running. That might explain why it has not been noticed before.

PR #662, opened 2023-02-26
Updated user documentation + calculate orbital period automatically so user doesn't have to specify it.

PeterBartram:


hannorein:
Can you elaborate a little on this? What prompted the change? 

Also looks like the unit tests are still failing...

PeterBartram:
The documentation update I promised back last year and finally around to doing. I noticed while updating this that the interface would be cleaner if users didn't have to specify the minimum dynamical timescale in the system for calculating the rectification period and that it was straight forward to make the change as I have to calculate all the periods anyway for the step size selection. 
Ultimately, I think it would be nicer if I periodically checked for the minimum period object in the system and updated the rectification period accordingly so that users would never have to specify a minimum period. I think that would require some more thorough testing though to make sure there aren't unexpected effects.

hannorein:
Cool. Thanks! I'll merge it in. Do you know if people have started using TES already?

PeterBartram:
Great, thanks! Unfortunately, I haven't heard of anyone using it yet.

PR #674, opened 2023-04-01
Adding ObliquityEvolution.ipynb to ipython_examples

michaelkmpoon:
Adding Obliquity Evolution tutorial, which reproduces Fig. 3 in Millholland+Laughlin2019 and Lu+2023.

Authored by Dang Pham + Michael Poon (equal contribution)


dtamayo:
This looks like a really nicely annotated example! I think the place for it would be in REBOUNDx rather than REBOUND. In order to run the automatic unit tests in a reasonable time, I think you would have to change the max times so that the whole notebook runs in ~20 seconds, but there could be text saying what times should be chosen to get capture etc. Of course this might be part of @hannorein 's course, in which case feel free to disregard!

hannorein:
It is part of my course. And your feedback is identical to mine ;-)

michaelkmpoon:
Thanks @dtamayo and @hannorein for the feedback! @dangcpham and I made changes and submitted a pull request to REBOUNDx. :D

PR #676, opened 2023-04-06
Submission of AST3100 A3

porpose:
Hi Hanno,

I've add a notebook for KL cycles as the submission of AST3100 A3.


porpose:
@hannorein I'm not sure my reply to the email is received; here are my thoughts for the correct-looking behavior despite too large timestep:

One possible explanation I can think of is that since KL cycle is a secular effect, it happens on a timescale much longer than the orbital period. In that case, although the large timestep makes the integrated orbital position inaccurate, it doesn’t really matter that much since we are only interested in the long-term variation of eccentricity and inclination. 

Another thing is I remember you mentioned that WHFast might be doing some automatic time-averaging while doing the integration. This probably makes the integration using WHFast more similar to secular approximation instead of a full N-body simulation, but I don’t know what’s the precise name of mechanism working there, if it actually exists.

PR #685, opened 2023-05-05
Fixed machine independence bug.

PeterBartram:
Issue was the use of the math.pow function. Replaced it with ias15 version of  x^1/7. 
Tested for bitwise energy violation matching for the longterm integration test is the unit test file (1e6 orbits for inner solar system) using a ubuntu desktop and a macbook M1.


hannorein:
Nice catch! This shouldn't matter for any practical purpose, but it will help a lot with the unit tests!

PR #687, opened 2023-05-08
New dt calculation for IAS15

dangcpham:
IAS15 old dt calculation method can be slow by giving very pessimistic time steps for certain problems. This new method calculates `dt` by comparing the second vs. third derivative, naturally giving a timestep:

$$ dt = \frac{y^{(2)}}{y^{(3)}} \cdot \zeta $$

where $\zeta \simeq 6 \times 10^{-2}$ is a constant that can be tuned (but not needed to be fine-tuned).

To use the new method, set `sim.ri_ias15.dt_mode=1`. The constant $\zeta$ can be set via `sim.ri_ias15.zeta`, with all the similar caveats as in changing `sim.ri_ias15.epsilon`.


dangcpham:
It is now possible to do input/output with `ri_ias15.dt_mode` and `ri_ias15.zeta`. Also, backward/forward integration now works (now comparing absolute values of $dt$).

src/rebound.h
```diff
@@ -577,6 +577,8 @@ enum REB_BINARY_FIELD_TYPE {
     REB_BINARY_FIELD_TYPE_IAS15_MINDT = 70,
     REB_BINARY_FIELD_TYPE_IAS15_EPSILONGLOBAL = 71,
     REB_BINARY_FIELD_TYPE_IAS15_ITERATIONSMAX = 72,
+    REB_BINARY_FIELD_TYPE_IAS15_DTMODE = 73,
```
dangcpham, line 580: I'm not quite sure if allocating 73 and 74 here for the new variables would incur any problem (i.e. backward compatibility with loading previous versions of REBOUND simulation archives)...

hannorein: You're right, this could be a problem (I doubt many would run into it). Choose a number higher than the highest one used.

dangcpham: It's now updated to 401 and 402.

hannorein:
Thanks. I've squashed merged this on the newias branch.

PR #697, opened 2023-06-22
Remove Python 2.7 dependency from workflow

dangcpham:
As of June 19, GitHub workflow no longer supports Python 2.7 by setup-python (https://github.com/actions/setup-python/issues/672). This pull request removes 2.7 dependency from the workflow.

Note: there is still a Python 2.7 ipynb dependency. Not sure how to work around this as it is there to test python 2.7 exclusively.

### DIFF
--- a/.github/workflows/ipynb.yml
+++ b/.github/workflows/ipynb.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["2.7"]
+        python-version: ["3.10"]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


hannorein:
Thanks. This seem good, but why does it still fail? I'm confused.

hannorein:
Oh, sorry. You mentioned it, the notebooks. So that needs to be updated too.

dangcpham:
All tests passed now. 

Do you still want to test for Python 2.7 in the future? It seems there are some hacks around to implement Python 2.7 in setup-python.

hannorein:
Excellent. Thanks!

I don't think there is a need to keep testing for 2.7...

Issue #705, opened 2023-07-10
clang-style C documentation comments would help with generated rust bindings


spenczar:
I am interesting in calling rebound (and assist) from Rust. To that end, I've generated Rust bindings to access `librebound` over a C FFI interface: https://github.com/spenczar/rebound-sys

This has gone pretty well!

One thing that hasn't worked is automatic conversion of documentation comments. It happens that the standard Rust tool for this job, `bindgen`, only is interested in doxygen-style comments: ones that start with `/**` or `///`.

It would be kind of nice if Rebound's headers used that style of comments. It's not a very big deal, though; these bindings are extremely low-level, and I'll certainly be writing a wrapper library that is much more human-friendly, so it's acceptable to just refer to `rebound.h`.

hannorein:
Many years ago, we had doxygen documentation for everything. But honestly, it was too much effort to maintain this specific syntax and just felt a bit inflexible. I suspect very few people look at this sort of documentation which is why we're now mostly using short examples to explain stuff. But I get your point. So what specifically does this bindgen tool want to see? Can you provide a few examples? Does it support the `/**<` style syntax (which I find the least annoying)? Does it need documentation for all the arguments and return values (which I find the most annoying)?

spenczar:
> So what specifically does this bindgen tool want to see?

I'm glad you asked this question, because I agree that Doxygen is obnoxious. Fortunately, bindgen doesn't actually require true Doxygen. It uses `clang` under the hood, so it wants the format used by `clang-doc`, which is a much more lightweight system:

A "documentation comment" is just any comment that starts with either `///`, or `/**`, or `/*!`. 

The contents of the "documentation comment" can be anything you like, but plain old text is best. Don't get cute with markdown or structured text inside. There's no special way you annotate that something is a input parameter or whatever, since the signature already has that information.

To annotate a declaration, the comment needs to be on the immediately previous line(s):
```c
/// foo does a thing
int foo(int x, int y);

/// bar does other things
struct bar {
   /// z is a useful field 
   int z;
};

/** this is a complicated function
 * so i'll give it a multiline
 * documentation comment
 */
void baz();
```

To annotate a struct field, the comment can be on the same line, too:
```c
/// bar does other things
struct bar {
  int z; /// z is a useful field
}
```

Enums need to be annotated with the comment in front of each value. For whatever reason, same-line annotations don't work:

```c
/// day of the week
enum WEEKDAY {
  /// the first day
  SUNDAY = 1;
[...]

PR #708, opened 2023-07-15
clang-docify a few things in rebound.h

spenczar:
As we discussed a bit in #705, clang-doc style documentation comments brings some benefits for automated tools, like generating bindings. This is a little mini-PR of a bit of what the changes might feel like. I tried to cover a bunch of the scenarios we'd encounter if we converted everything.

If I'm honest, I don't think these seem like improvements for the C code. Enums get _way_ bigger to document every value. Inline function comments need to be put in front, which kind of chops things up and makes symmetries less obvious. The only clear success case is probably commit 5df64af, which I think certainly makes things a little more readable.

### DIFF
--- a/src/rebound.h
+++ b/src/rebound.h
@@ -1100,19 +1100,27 @@ void reb_free_ode(struct reb_ode* ode);
 // Miscellaneous functions
 uint32_t reb_hash(const char* str);
 double reb_tools_mod2pi(double f);
-double reb_tools_M_to_f(double e, double M); // True anomaly for a given eccentricity and mean anomaly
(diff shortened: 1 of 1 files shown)


hannorein:
I agree with everything you wrote. I'm not very enthusiastic about the changes except maybe a few like the 5df64af you point out. In most cases they make the code less readable - but I'm also obviously biased because I know it from before. If you were to push for it I think you could convince me to go ahead. But if you're not, then I'd say we don't do it?

spenczar:
I agree, this doesn't seem worth it. Let's drop it - I'll add notes to the bindings directing any users to look up documentation by looking at `rebound.h`.

PR #727, opened 2023-10-28
Triggerrebx

dtamayo:
I find it hilarious that this pull request amounts to half a line of code per commit 🤦‍♂️


dtamayo:
Please squash merge if things look OK

hannorein:
This workflow only gets triggered manually? Is that intended?

dtamayo:
Oops, thanks for catching that! Now dropped below half a line of code per commit 😄

PR #734, opened 2023-12-01
Fixed documentation link issue.

zyrxvo:
There is a link in the current documentation listed as Examples -> Planetary Systems -> Just getting started -> None. As far as I can tell, this fixes the link which goes to "c_examples/reb_simulation_add_fmt.md" resulting in a 404 error but seems intended for "c_examples/reb_add_fmt/".


hannorein:
Thanks.

PR #738, opened 2024-01-12
added repr to orbits and fixed orbitalelements.ipynb

dtamayo:
Fixed OrbitalElements.ipynb to new syntax of sim.particles[i].orbit().

Also added a __repr__ to the Orbit class so when we finish a cell with e.g.

sim.particles[1].orbit()

it will give us useful information instead of just Orbit object at memory loc X


hannorein:
Thanks! Couldn't `__repr__` just call `__str__`? I see there is a conceptual difference between the two but we're not making you of it right now. (We could in principle, `__str__` could output something like "eccentric orbit" or "hyperbolic orbit" to make it more human-interpretable)

hannorein:
makes sense!

(Don't know why the REBOUNDx trigger fails)

dtamayo:
emailed you

PR #745, opened 2024-01-24
updated horizons

dtamayo:
Was messing around with this for the class I'm teaching this semester, and noticed that a couple things were out of date from what I wrote a long time ago. All I did was change the text

- Looking at the code we're passing hours minutes and seconds to horizons (so removed line about rounding things to the nearest minute)

- I tested that the values from horizons use the solar system barycenter as the origin, so fixed the last section to say that


hannorein:
Thanks!

PR #746, opened 2024-01-25
Collision Test Fix

tigerchenlu98:
Hey Hanno, this one was quite tricky but think I figured it out!

1) Some issues with the scope of nbody_ode. After a merge (and seemingly only after a merge), only the local variable nbody_ode is updated, not the global r->odes[0]. To be honest I don't entirely understand why your code didn't work, everything looked fine to me - but I reassign y within the loop now and it seems to work.

2) Unrelated bug, but realized over the course of debugging that it's important to still update the actual r->t variable during the bs step. collision_resolve checks the last time of collision against the current simulation time, so this fails if the collision happens during the first call of bs_step.

### DIFF
--- a/src/integrator_trace.c
+++ b/src/integrator_trace.c
@@ -351,26 +351,28 @@ void reb_integrator_trace_bs_step(struct reb_simulation* const r, double dt){
     // run
     const double old_dt = r->dt;
     const double old_t = r->t;
-    double t = r->t;
+    //double t = r->t;
     const double t_needed = r->t + dt;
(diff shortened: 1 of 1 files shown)


hannorein:
Amazing. Thank you! I don't understand it fully yet... will look at it tomorrow.

hannorein:
Very well spotted! The issue was that when BS accepts a step, it overwrites the initial y array by swapping a pointer, rather than copying all the individual values (because it's faster). And yes, to the other issue regarding the time during collisions! Thank you!

PR #762, opened 2024-03-21
PRS Pericenter Condition

tigerchenlu98:
Forgot to put in the PR yesterday, but this has the PRS23 condition with eta=1 as the default.


hannorein:
Thanks for that! 

There are so many white space changes ... they make it hard to see what actually changed (not much!). Any chance you could get rid of them, especially in the rebound.h and rebound.c files? See e.g. https://stackoverflow.com/questions/13793241/make-git-undo-any-whitespace-only-changes If this git hackery gives your problems, let me know and I'll do it .. but then you'll need to pull from me afterwards to stay in sync.

Hanno

tigerchenlu98:
Done!

PR #771, opened 2024-05-05
TRACE Main PR

tigerchenlu98:
Hi Hanno,

Here's the hopefully final PR! Things look pretty clean (except the notebooks, don't think there's a way around that). I replaced all the examples, the only one which is doing worse is PrimordialEarth - not sure why, I need to look into it more.

Let me know if everything looks good!

Tiger


hannorein:
Wow! This looks great! I looked over it just now and only found very minor things. I want to stare at it a little longer, just to make sure. But I think we're basically done!

docs/integrators.md
```diff
@@ -400,9 +400,158 @@ The `reb_integrator_mercurius` structure contains the configuration and data str
...
+`double (*S) (const struct reb_simulation* const r, const unsigned int i, const unsigned int j)`
+:   This is a function pointer to the switching function for close encounters between non-central bodies.
+    If NULL (the default), the default switching function will be used.
+    The arguments `i` and `j` are the indices of the two particles considered.
+    The return value is a scalar.
+    If this function returns a negative value, it means a close encounter has been flagged.
+
+    - Default switching function
+
+        This is a similar (but slightly modified) switching function used in MERCURY and MERCURIUS.
```
hannorein, line 436: This needs to be a bit expanded. What is the default switching function?

docs/integrators.md
```diff
@@ -400,9 +400,158 @@ The `reb_integrator_mercurius` structure contains the configuration and data str
...
+        ```python
+        sim = rebound.Simulation()
+        sim.ri_trace.S = "default"
+        ```
+        
+`double (*S_peri) (const struct reb_simulation* const r, const unsigned int j)`
+:   This is a function pointer to the switching function for close encounters involving the central body.
+    If NULL (the default), the default switching function will be used.
+    The argument `j` is the index of the non-central particle considered.
+    The return value is a scalar.
```
hannorein, line 460: For both switching functions' return value, we seem to use 0/1 in the code, but the documentation says pos/neg. Need to make this consistent.

hannorein:
Please merge in my branch `tigerchenlu98-newTRACE4_FULL_fbs` for some minor changes/cleanup.

Then see the two small queries above.

tigerchenlu98:
Both done!

docs/integrators.md
tigerchenlu98: Don't know how much detail we want to go into in the docs, but how does it look now?

hannorein:
I pushed a few updates to the notebooks (you can merge my `tigerchenlu98-newTRACE4_FULL_fbs` branch again). Mainly typos, consistency things. 

Not sure why the windows tests sometimes (?) fail. Does not seem related to TRACE.

We either need to revert the primordial earth example to use mercurius, or update it (it still mentions Mercurius), depending on what your tests show.

PR #772, opened 2024-05-07
Missing Inertial<->DHC transform in TRACE

tigerchenlu98:
There was a conversion between DHC <-> Inertial for the FULL PERI prescriptions that somehow got lost in some of the updates: this was making the violent system behave very poorly. I've corrected it and re-checked the relevant examples, and everything seems to be working as expected again. 

The unit tests didn't catch this because the pericenter test was commented out. I've added it back in and confirmed that it fails if the transformation is not there, and passes otherwise.

Also fixed a typo in the documentation :)


hannorein:
Thanks for catching that.

PR #781, opened 2024-07-07
Fixed issue where minor and patch numbers were not referenced correctly with self.

zyrxvo:
This resulted in a crashing error when trying to display the details of a SimulationArchive in ipython and jupyter notebook.


hannorein:
Thanks

PR #787, opened 2024-07-17
Remove unused modules

AlexKurek:


hannorein:
Thanks. This is much appreciated. Did you check each import manually or did you use some script?

AlexKurek:
> Thanks. This is much appreciated. Did you check each import manually or did you use some script?

I used Pylance in VS Code.

PR #789, opened 2024-07-24
Remove unused things, applied suggestions from Ruff

AlexKurek:


hannorein:
Thanks. Some of the REBOUND code base could definitely be cleaned up. But I don't think these changes are very helpful, nor am I sure really not changing the behaviour in every possible case. So I think I'll close this for now.

Issue #794, opened 2024-08-24
[TRACE+] Binary files size from simulation archive dependent on integrator


gabrieltxg:
**Environment**
Which version of REBOUND are you using and on what operating system?
 - REBOUND Version: 4.4.1
 - API interface: Python
 - Operating System (including version): Linux 

**Describe the bug**
Actually I'm not sure if it is a bug, but just like the title, the final file size of the binaries saved depend on the integrator chosen. 

I was benchmarking and trying to reproduce some benchmarks comparing IAS15, WHfast, MERCURIUS and TRACE (with different pericenter settings).

For IAS15 and TRACE+FULLIAS15 the final file size is almost 10 times bigger, while for WHfast it's around 30% bigger.

Same sampling and same system (with initial conditions slightly different).

Does it have something to do with with the number of internal timesteps? 

**To Reproduce**
I'm trying to reproduce Lu et al. (2024) papers, specifically the "violent systems" section, since i'm interested in compact planetary systems.

Some initial conditions are different but the problem persisted for different setups.

Here's my code: 

```
import rebound
import numpy as np
import matplotlib.pyplot as plt
import pandas as pd
import random
import time

def hill_radii_dist(ai, ms, mi, mj, delta):
    eta = ((mi + mj)/(3*ms))**(1/3)
    aj = ai*(2+delta*eta)/(2-delta*eta)
    return aj

def gaussian(mu, sigma):
    u1 = np.random.uniform(size=1)
    u2 = np.random.uniform(size=1)
    R_sq = -2 * np.log(u1) * sigma ** 2
    theta = 2 * np.pi * u2
    z1 = np.sqrt(abs(R_sq)) * np.cos(theta) + mu
    z2 = np.sqrt(abs(R_sq)) * np.sin(theta) + mu
    return z1

def simulation_delta(delta, n, e = 0, i = 0, sig = 0):
    # Planets semimajor axis are drawn from a graussian distribution with kmean and kstd as mean 
    # orbital separation parameter and spreading on this separation.
    kmean = delta #Initial orbtal separation between two adjacent planets, in terms of mutual hill radius
    kstd = sig #Standard deviation 

    sim = rebound.Simulation()
    msun = 1 #Msun
    m_earth = 1e-4 ## 10% of jupiter mass
    sim.add(m = msun)
    ai = 5 #Semimajor axis of the innermost planet
    m1= m_earth*np.random.uniform(3,9) # Mass is drawn from a uniform distribution
    #############3
[...]

PR #795, opened 2024-08-27
Reset BS/IAS15 in TRACE to reduce file size

hannorein:
It would be nice to reduce the file size in TRACE integration. 

@tigerchenlu98, I naively thought these changes should not have any effects on the actual integration. But the units tests are failing. Do you understand why?

### DIFF
--- a/src/integrator_trace.c
+++ b/src/integrator_trace.c
@@ -530,6 +530,9 @@ void reb_integrator_trace_bs_step(struct reb_simulation* const r, double dt){
         r->N_odes = N_odes_backup;
 
         r->t = old_t;
+        
+        // Resetting BS here reduces binary file size.
+        reb_integrator_bs_reset(r);
     }
 }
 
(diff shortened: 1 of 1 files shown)


hannorein:
Sorry, my bad: the unit tests pass! Do you see any downside coming from these changes? There should only be a negligible overhead in terms of runtime because the reset functions were already called at the beginning and calling them twice doesn't take long. But you have all these timing tests set up, so maybe you can verify that this is indeed the case?

tigerchenlu98:
Hey Hanno,

I see no reason the actual integration should change. 

In terms of overhead, the only thing I'm concerned about is reallocing the nbody ode every time we call reset. I'm not sure if this would actually be an issue, but it's easily tested -- I'll run them and get back to you.

tigerchenlu98:
I ran the tests for all three pericenter modes -- integrations were all completely identical and runtimes were not meaningfully longer (less than 1 percent), so I think this is good to go!

hannorein:
We already call reset at the beginning of the step, and then reallocating everything, so I don't think there's a significant overhead. And your tests seem to confirm that! Thanks! I'll merge it in...

hannorein:
Fixes #794

PR #804, opened 2024-11-08
Implementation of Barycentric Coordinates

luke-wriglesworth:
For the transformations I used something similar to Jacobi coordinates, where the first coordinate points to the center of mass and the other coordinates are calculated relative to it.


luke-wriglesworth:
Hi Hanno, 

I think it is ready to be merged now. There is a test failing because of the codecov token, I'm not sure why.

hannorein:
Thanks Nick. This looks great. Give me a couple of days to test various edge cases. In the meantime, maybe you can think about examples where we can get good performance using these coordinates.

hannorein:
Merged manually (after squashing).

PR #806, opened 2024-11-22
Fix for transformation unit tests

luke-wriglesworth:
Fixes typo in transformation unit tests which caused only the first particle to be checked.


hannorein:
Thank you!

PR #822, opened 2025-03-13
Add symplectic correctors for barycentric coordinates.

luke-wriglesworth:
I tested this with the outer solar system and the energy error is lower than WHDS and Democratic Heliocentric coordinates. It is also lower than Jacobi if correctors aren't used.


hannorein:
Thanks! This looks great.

PR #825, opened 2025-04-01
TRACE Fragmentation minus safe mode

tigerchenlu98:
Here's the new PR! Everything in the other branch, minus safe mode -- we can either close that one, or keep it around for the new additions.


hannorein:
Thanks!

The fragmentation example has `r->ri_trace.r_crit_hill = 10.;`. I'm also currently running the mercurius tests with `r->ri_trace.r_crit_hill = 10.;`. I suspect I'm not doing that right? Can you confirm what are the radii that I should use for Mercurius and Trace respectively?

hannorein:
Never mind! I think it all makes sense... the setting of 10 only affects TRACE...

tigerchenlu98:
Exactly! Nothing has changed with the MERCURIUS settings by the way -- if you've already started those, no need to change anything!

tigerchenlu98:
I'm happy with it! A few checks still fail -- most of these are from the fragmentation example, but the "Checking test coverge" check has been failing the whole time. If that's not a concern, I think we're ready.

hannorein:
Cool (The coverage test just fails because the credentials are missing on the pull requests.)

hannorein:
Merged in 6746b892346df6ac52281ab6a16a3e87dad3f94a

PR #828, opened 2025-04-18
MERCURIUS switching condition in TRACE

tigerchenlu98:
This basically implements the MERCURIUS dcrit calculation in TRACE! It only depends on v^2, so it should be time-reversible -- tested this with the violent system and it was totally fine. I feel like this is probably the best way to directly compare TRACE and MERCURIUS. We can think a little harder about if we actually want this to be the default switching condition in TRACE, but for now I think it's the best way to run the simulation in the Note. I've run a quick test case for a day and we are catching roughly the same number of close encounters with TRACE and MERCURIUS.

@hannorein, I have also added the problem.c file with all necessary inputs to this commit, and CAREFULLY commented out all the printf/fprintf statements. Could you toss up a few of these runs onto your cluster?


hannorein:
Quickly question before I run these: Currently, `move_to_com()` gets called every timestep in the heartbeat function. Is that there for a reason?

tigerchenlu98:
hmm, it's not clear to me. @haniyehtajer, do you have any insight? This heartbeat function is not in the original Childs & Steffan code. Perhaps it's a typo, and can be put inside the braces -- we would need to reset to CoM after removing a particle. Or perhaps the collision prescription messes with the CoM in some way, and it was easiest to put it here rather than any of the collision prescriptions?

Would resetting to the CoM each timestep be a significant computational burden? There's no downside, right?

haniyehtajer:
I added that line because the system seemed to drift away in several of my simulations. So, doing this did fix the problem, but I'm not sure if there is a better way to do this.

hannorein:
Should I close this now?

Issue #826, opened 2025-04-03
peri_mode not in Simulationarchive


hannorein:
The peri_mode flag of TRACE is currently not stored in the binary files.

PR #830, opened 2025-04-19
Force accept fix

### LINKED ISSUE #829: Check TRACE force accept logic
I might be missing something...

We force accept a step if `N_allocated_collisions` is bigger than 0. I'm not sure this ever gets reset to 0 after there has been a collisions. This would mean we force accept all steps after one collision occurred.

tigerchenlu98:
Fixes #829 and doesn't appear to be breaking anything.

Just checking -- `collisions_N` only tracks collisions over one timestep, so there's no need to include it in `output.c` correct?

### DIFF
--- a/rebound/simulation.py
+++ b/rebound/simulation.py
@@ -1543,6 +1543,7 @@ Simulation._fields_ = [
                 ("collision_resolve_keep_sorted", c_int),
                 ("collisions", c_void_p),
                 ("N_allocated_collisions", c_int),
+                ("collisions_N", c_uint),
                 ("minimum_collision_velocity", c_double),
                 ("collisions_plog", c_double),
                 ("max_radius", c_double*2),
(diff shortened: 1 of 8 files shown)


hannorein:
Yes. No need to store it. FWIW I’d set collisions_N=0 at the beginning of the collision search. Then the variable always reflects the number of collisions found in the last collision search, regardless of how the collision search was initiated.

hannorein:
Thanks. I’ll merge it on Monday. Any idea if it makes a difference?

tigerchenlu98:
Tough to say with our usual examples -- the violent example, for instance, was working just fine with the bug. But my thoughts are that those examples are run for a relatively short time with few step rejections anyways, so perhaps the bug does not materialize.

I've started a few test run with the fragmentation problem, which is probably a good test for this. By Monday, they should have run enough to see if there are significant differences.

By the way, this also fixes #826

hannorein:
Thanks! Merged. I changed a few more things since adding the variable in the rebound structure allows us to get rid of the local one as well as the various pointers passing it along in functions.

PR #849, opened 2025-07-15
check for convergence on M_to_E

dangcpham:
The following setup fails to converge when executing the function `reb_M_to_E` in the `src/tools.c` file. I implemented a simple check to throw a warning message when convergence is not reached after 100 iterations. Perhaps a better implementation is to perform a different method to solve Kepler's equations when Newton's method fails.

```
#include "rebound.h"

int main(int argc, char **argv) {
    double e = 0.1;
    double M = 0.991;
    double E = reb_M_to_E(e, M);

    return 0;
}
```

### DIFF
--- a/src/tools.c
+++ b/src/tools.c
@@ -568,16 +568,24 @@ double reb_mod2pi(double f){
 
 double reb_M_to_E(double e, double M){
     double E;
-    int converged = 0;
-    if(e < 1.){
-        M = reb_mod2pi(M); // avoid numerical artefacts for negative numbers
-        E = e < 0.8 ? M : M_PI;
-        double F = E - e*sin(E) - M;
+    double F;
 
+    int converged = 0;
+    if (e < 1.){
+        M = reb_mod2pi(M); // avoid numerical artefacts for negative numbers
+
+        // E = e < 0.8 ? M : M_PI;
+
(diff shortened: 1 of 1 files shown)


dangcpham:
Here is a plot showing where the function fails to converge (black dots). The red line shows the mean anomaly (M=0.991) in the first comment.

M_to_E_convergence_100

I also tried to increase the number of iterations to 1000, but the plot looks identical:

M_to_E_convergence_1000

hannorein:
Thanks Dang! I’ll try to implement a better version!

dangcpham:
I found the the original condition `fabs(F) < 1.e-16` is actually a bit too stringent since it is very close to floating point precision. I reduced the criteria to ``fabs(F) < 1.e-16`, which allows convergence for all cases tested.

In addition, I also implemented a different initial guess based on [Napier (2024)](https://arxiv.org/abs/2411.15374) and [Danby & Burkadt (1983)](https://link.springer.com/article/10.1007/BF01686811). Here are some comparison between the new initial guess and the previous REBOUND implementation:

In general, the Napier (2024) guess converges faster:

image

image

which improves the runtime as can be seen in this histogram:

image

and this plot showing the regions where the new guess is faster than original:

image

hannorein:
Thank you!

dtamayo:
Looks awesome!

PR #857, opened 2025-07-31
updated J2 sign and added comparison to analytics

dtamayo:
* updated sign
* removed obliquity (assumes spin along z axis)
* added back reaction
* added reference to reboundx in description
* added comparison to orbit-averaged analytic expressions (that match!)


hannorein:
Amazing! Thank you!

PR #871, opened 2025-12-27
fix race conditions in Kepler step for WHFast (closes #870)

### LINKED ISSUE #870: Possible race condition in WHFast Kepler step
**Environment**

 - REBOUND Version: 4.4.11 (also latest commit on the main branch)
 - API interface: C
 - Operating System (including version): MacOS Tahoe Version 26.1

**Describe the bug**
Possibly a race condition when in `reb_whfast_kepler_step` of WHFast.

**To Reproduce**
I just realized this while reading the code. In the file `integrator_whfast.c` we have the function

```C
void reb_whfast_kepler_step(const struct reb_simulation* const r, const double _dt){
    const double m0 = r->particles[0].m;
    const double G = r->G;
    const unsigned int N_real = r->N-r->N_var;
    const int N_active = (r->N_active==-1 || r->testparticle_type ==1)?(int)N_real:r->N_active;
    const int coordinates = r->ri_whfast.coordinates;
    struct reb_particle* const p_j = r->ri_whfast.p_jh;
    double eta = m0;
    switch (coordinates){
        case REB_WHFAST_COORDINATES_JACOBI:
#pragma omp parallel for 
            for (int i=1;i<(int)N_real;i++){
                if (i<N_active){
                    eta += p_j[i].m; // <--- Possible race condition here?
                }
                reb_whfast_kepler_solver(r, p_j, eta*G, i, _dt);
            }
            break;
        case REB_WHFAST_COORDINATES_DEMOCRATICHELIOCENTRIC:
#pragma omp parallel for 
            for (unsigned int i=1;i<N_real;i++){
                reb_whfast_kepler_solver(r, p_j, eta*G, i, _dt); //  eta = m0
            }
            break;
        case REB_WHFAST_COORDINATES_WHDS:
#pragma omp parallel for 
            for (int i=1;i<(int)N_real;i++){
                if (i<N_active){
                    eta = m0+p_j[i].m;
                }else{
                    eta = m0;
                }
                reb_whfast_kepler_solver(r, p_j, eta*G, i, _dt);
            }
            break;
        case REB_WHFAST_COORDINATES_BARYCENTRIC:
            eta = p_j[0].m;
            for (unsigned int i=1;i<(int)N_real;i++){
                reb_whfast_kepler_solver(r, p_j, eta*G, i, _dt);
            }
            break;
    };
}
```

Here in the `REB_WHFAST_COORDINATES_JACOBI` case we are mutating `eta` in a shared parallel context. In the `REB_WHFAST_COORDINATES_WHDS` case perhaps `eta` needs to be private to the OMP directive? I tried to make sense of by referring to [Hernandez & Dehnen (2017)](https://academic.oup.com/mnras/article/468/3/2614/3061366)'s Equation (36a) for Jacobi co-ordinates. Perhaps the right way to do this is having two loops and have an array for `eta` which stores the cumulative masses, although I am not really sure. Any sort of help/clarification would help!

**Additional context**
None

**Name/Affiliation**
Maitrey, NISER Bhubaneswar

kmaitreys:
Please have a look. Although, there is no race condition now, I am just concerned if I have followed the algorithm correctly (in case of Jacobi co-ordinates) from the paper, so please let me know if there are problems in it.

I had one question, is there any particular example we should now test for regressions (if any) and physical correctness as well? In the examples directory, I think none of WHFast examples are compiled with OpenMP and I was not sure which one would be best to test the effect of this change.

Please feel free to merge or comment on this when you get time, after holidays.

### DIFF
--- a/src/integrator_whfast.c
+++ b/src/integrator_whfast.c
@@ -509,15 +509,21 @@ void reb_whfast_kepler_step(const struct reb_simulation* const r, const double _
     struct reb_particle* const p_j = r->ri_whfast.p_jh;
     double eta = m0;
     switch (coordinates){
-        case REB_WHFAST_COORDINATES_JACOBI:
+        case REB_WHFAST_COORDINATES_JACOBI:{
+            double eta_jacobi[N_real];
+            eta_jacobi[0] = m0;
+            for (int i=1; i<(int)N_real;i++){
+                eta_jacobi[i] = eta_jacobi[i-1];
+                if (i < N_active){
+                    eta_jacobi[i] += p_j[i].m;
+                }
+            }
 #pragma omp parallel for 
             for (int i=1;i<(int)N_real;i++){
-                if (i<N_active){
-                    eta += p_j[i].m;
-                }
-                reb_whfast_kepler_solver(r, p_j, eta*G, i, _dt);
+                reb_whfast_kepler_solver(r, p_j, eta_jacobi[i]*G, i, _dt);
             }
             break;
+        }
         case REB_WHFAST_COORDINATES_DEMOCRATICHELIOCENTRIC:
 #pragma omp parallel for 
             for (unsigned int i=1;i<N_real;i++){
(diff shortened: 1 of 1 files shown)


hannorein:
There are a couple of issues with your fix so I'll close this and reimplement it slightly differently.

PR #896, opened 2026-03-24
TRACE WB Coordinates

tigerchenlu98:


tigerchenlu98:
Hey @hannorein, believe this should have incorporated all of our suggestions except the one about moving gravity terms from update_accelerations to integrator_trace -- don't think I understand what you mean there, could you elaborate?

hannorein:
I can help move everything into the integrator_trace.c file at a later point. But maybe you can use `REB_GRAVITY_TRACE` for both coordinates? Just use an if statement in case REB_GRAVITY_TRACE. That would help down the road. Thanks.

tigerchenlu98:
taken care of. Don't merge just yet though! A pesky bug with collisions w/ the central star that I'm trying to work out... will ping you when I've resolved it

tigerchenlu98:
@hannorein put in some documentation! All the unit tests pass fine on my machine... not quite sure what's going on on the other architectures, it's all order unity issues.

hannorein:
The speed test just needs to be relaxed as any timing is very inconsistent on the CI runner. I'm not sure about the others. It's strange that they fail on some architectures but not on others. Hopefully this isn't an indication of some sort of undefined behaviour...

hannorein:
There are some merge conflicts left.

tigerchenlu98:
done!

PR #900, opened 2026-04-02
Feat: Add try_create API to prevent Segfault on memory allocation failure

yuanzui-cf:
Hi! I'm currently working on a Rust wrapper for REBOUND. During development, I noticed that `reb_simulation_create` calls `memset` immediately after `calloc` without checking for a `NULL` return value. If the memory allocation fails, this triggers an immediate Segmentation Fault.

For high-level language wrappers (like Rust, Python, etc.), it's crucial to catch these Out-Of-Memory (OOM) errors gracefully rather than having the entire process hard-crash.

## Changes

To fix this without breaking backward compatibility, I implemented the following:
- Add `reb_simulation_try_create()`, which returns `NULL` on allocation failure
- make `reb_simulation_create()` call `reb_simulation_try_create()` and fail explicitly on allocation failure

## Notes

While reviewing the code, I noticed there are many other `malloc` and `calloc` calls throughout the codebase that also lack `NULL` checks. To keep this PR focused and easy to review, I only addressed the primary simulation initialization. However, it might be worth adding similar checks to other memory allocations in the future to further harden the library.

Thanks for the great library!

### DIFF
--- a/src/rebound.c
+++ b/src/rebound.c
@@ -380,11 +380,17 @@ int reb_simulation_reset_function_pointers(struct reb_simulation* const r){
 }
 
 struct reb_simulation* reb_simulation_create(){
-    struct reb_simulation* r = calloc(1,sizeof(struct reb_simulation));
-    reb_simulation_init(r);
+    struct reb_simulation* r = reb_simulation_try_create();
+    if (r == NULL) reb_exit("Memory allocation failed.");
     return r;
 }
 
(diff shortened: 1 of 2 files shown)


hannorein:
I'm sorry, but this is not useful and simply bloats the code. It's the kind of thing an LLM would suggest. There is no practical scenario where memory allocation might fails.

PR #909, opened 2026-05-12
Added MPI communication header file to integrator

ConnorA19:
Added mpi communication header file to integrator for mpi functionality that was changed in e853676


hannorein:
Thanks. The MPI features desperately needs some TLC.

PR #916, opened 2026-06-06
freeze already converged lanes

Rishit-dagli:
kepler convergence checks slowest converging lane and updates all lanes. this change makes it not update particles already converged.

### DIFF
--- a/src/integrator_asm512.s
+++ b/src/integrator_asm512.s
@@ -276,10 +276,14 @@
     halley
   
     movq            $0, %rcx                    # Newton loop counter
-.NewtonLoop\@:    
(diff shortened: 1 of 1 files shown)


hannorein:
Thanks! This looks good. Just one question: Rather than restoring the old X value for converged lanes, have you thought about not updating the X values anymore after they are converged? Wouldn't that allow us to avoid the `vmovapd` instruction? I might be missing something...

Rishit-dagli:
yeah, i did this because it was very little change needed

but got it, now we do a write only updates the lanes whose mask bit is 1 for all the four writes newton does in https://github.com/hannorein/rebound/pull/916/commits/29b9d6cfcc044c1af72cbf00e9ad6ed3ddcf7dd2

PR #917, opened 2026-06-06
change build to use avx when available

Rishit-dagli:
for the build system i have pretty much tried to do a similar approach as: https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/x86/cpu.c and https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/x86/float_dsp_init.c. for the python i have largely tried to take a lot from what numpy did: https://numpy.org/devdocs/building/cpu_simd.html.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@
 # All project metadata lives in pyproject.toml.
 
 from setuptools import setup, Extension
+from setuptools.command.build_ext import build_ext
 from glob import glob
 import os
 import sys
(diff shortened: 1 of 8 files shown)


hannorein:
This also looks great! Thank you. With these changes, I guess we could get rid of the AVX512 precompiler macro now?

Rishit-dagli:
doen in https://github.com/hannorein/rebound/pull/917/changes/25fa408b05d34013d0310f65703aa2b71b52b384

PR #918, opened 2026-06-07
allow using lower number of planets in whfast512

Rishit-dagli:
i tested on 1,2,4, systems with 8,4,2 planets. we get the same outputs if we simulated the system by itself or with 2 or 4 copies of it as applicable. i also tested with 3,5,6,7 planets.


hannorein:
One thing that still needs to be addressed is the coordinate transformations for less than 8 planets.

PR #924, opened 2026-07-05
add missing python code block opening tag in timestepping.md

SharpIceX:
I was going through the docs and noticed the `sim.stop()` code block on the timestepping page was missing its opening tag, so I fixed it while I was at it.


hannorein:
Thanks.

Issue #927, opened 2026-07-16
Documentation of value ranges of angular orbital elements


bischoff-m:
Hello everyone,

my code uses the orbital elements that rebound produces and relies on the specific definition ranges for data validation. The only mention I could find of the ranges for the angles in the documentation is the [IPython example titled Orbital Elements](https://rebound.hanno-rein.de/ipython_examples/OrbitalElements/). There it states:

> REBOUND will always output angles in the range $[-\pi, \pi]$, except the inclination which is always in $[0, \pi]$.

When I got some validation errors, I checked with this script:

```py
import numpy as np
import rebound

print(f"rebound version: {rebound.__version__}")

sim = rebound.Simulation()
sim.add(m=1.0)

rng = np.random.default_rng(0)
n = 10000

low, high = -10 * np.pi, 10 * np.pi
e = rng.uniform(0, 0.9, n)
inc = rng.uniform(low, high, n)
Omega = rng.uniform(low, high, n)
omega = rng.uniform(low, high, n)
f = rng.uniform(low, high, n)

for i in range(n):
    sim.add(
        m=0.0,
        a=1.0,
        e=e[i],
        inc=inc[i],
        Omega=Omega[i],
        omega=omega[i],
        f=f[i],
    )

orbits = sim.orbits()

for name in ["inc", "Omega", "omega", "pomega", "theta", "f", "M", "E", "l"]:
    values = np.array([getattr(o, name) for o in orbits])
    print(f"{name:8s} min={values.min():.6f}  max={values.max():.6f}")
```

The output is:

```
rebound version: 5.0.0
inc      min=0.000420  max=3.141585
Omega    min=-3.141110  max=3.141554
omega    min=0.001524  max=6.281596
pomega   min=-6.209577  max=6.278367
theta    min=0.000195  max=6.282877
f        min=0.000419  max=6.282324
M        min=0.000127  max=6.283137
E        min=0.000246  max=6.282898
l        min=0.000381  max=6.282764
```

Judging from this, the definition ranges are
- $[0, \pi]$ for the inclination,
- $[-\pi, \pi]$ for the longitude of ascending node,
- $[-2\pi, 2\pi]$ for the longitude of pericenter, and
- $[0, 2\pi]$ for all other angles.

Are these ranges correct? Specifically, are they e.g. $\leq 2\pi$ or $< 2\pi$?

[...]

PR #930, opened 2026-08-07
Move Omega and pomega to [0, 2pi) range and update docs

bischoff-m:
This pull request adjusts the definition ranges for angles for consistency. 

Changes made:
- Wrap Omega and pomega to $[0, 2\pi)$ in [`reb_orbit_from_particle_err`](https://github.com/hannorein/rebound/blob/main/src/tools.c#L1190) using `reb_mod2pi`.
- Update docs to explicitly state angle ranges.

The updated ranges are now
- $[0, \pi]$ for the inclination and
- $[0, 2\pi)$ for all other angles.

See [this issue](https://github.com/hannorein/rebound/issues/927).


hannorein:
Great! Thank you very much!

PR #931, opened 2026-08-18
Patch

dtamayo:
minor memory leaks and edge cases


hannorein:
Thanks for catching these!

PR #933, opened 2026-09-10
consistent newton

Rishit-dagli:


Rishit-dagli:
merge this after #932 because this commits on top of commits in that branch

Issue #934, opened 2026-09-10
Low precession for Mean anomaly


hannorein:
As reported by @chrisormel, the mean anomaly can have low precision in some cases. Here is a short example demonstrating this:

```python
import rebound
import numpy as np
import matplotlib.pyplot as plt
Ms = np.logspace(-16,0,100)
Ms_out = np.zeros(len(Ms))
for i, M in enumerate(Ms):
    sim = rebound.Simulation()
    sim.add(m=1)
    sim.add(e=0.1, a=1, M=M)
    Ms_out[i] = sim.orbits()[0].M
fig, ax = plt.subplots()
ax.set_xscale("log")
ax.set_yscale("log")
ax.plot(Ms, Ms_out)
```

Which produces this plot:

<img width="559" height="417" alt="Image" src="https://github.com/user-attachments/assets/ac61a48b-2900-4029-bc9b-68e3faee4268" />

hannorein:
aaef623c should fix this. It passes all units tests. The above plot now looks like this:

<img width="565" height="417" alt="Image" src="https://github.com/user-attachments/assets/8f1cc44b-4e12-47c3-a2ad-ae08f99c2ff7" />

PR #936, opened 2026-09-14
Improve eccentric anomaly precision

hannorein:
See #934


hannorein:
@chrisormel If you have a minute, let me know if this fixes all your issues! Thanks

